PR #24, opened 2020-01-31
Phoenix dockerized deterministic build

araspitzu:
This PR adds a docker image (via Dockerfile) that can be used to build phoenix without installing android studio or the android sdk dependencies, the build is also deterministic. Note that you need to build and publish locally `eclair-core` for this to work.


sstone:
I tested this and can generate the exact same release apks on 2 different machines. Debug apks are different but it's not an issue as we won't release them.

PR #30, opened 2020-03-04
Add italian translation

araspitzu:
Add the italian translation strings, I've kept the most common word in english such as _wallet_ or _seed_ and also some technical terms like _block height_ or _payment hash_. Below a few screenshots of how it looks like:

image
image
image
image
image
image
image
image

### DIFF
--- a/app/src/main/res/values-it/strings.xml
+++ b/app/src/main/res/values-it/strings.xml
@@ -1,0 +1,492 @@
+<!--
+  ~ Copyright 2019 ACINQ SAS
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<resources>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-it/strings.xml
```diff
@@ -0,0 +1,492 @@
...
+  <string name="pindialog_error_donotmatch">PIN non corretto.</string>
+
+  <!-- //////////////// Biometric dialog //////////////////// -->
+
+  <string name="biometricprompt_title">Wallet bloccato</string>
+  <string name="biometricprompt_negative">Sblocca con il PIN</string>
+
+  <!-- //////////////// placeholders ////////////////-->
+
+  <string name="lipsum_short">Etiam porttitor egestas faucibus. Curabitur condimentum eros non ipsum elementum egestas.</string>
```
dpad85: There are some texts like these placeholders that should not be translated.

The should be removed from the translated file, and in the original default `strings.xml`, these fields must have the `translatable="false"` tag.

app/src/main/res/values-it/strings.xml
```diff
@@ -0,0 +1,492 @@
...
+  <string name="newseed_words_td"><![CDATA[<td><font color="rgba(100,100,100,.3)"><small>#%1$d</small></font> <b>%2$s</b></td>]]></string>
+  <string name="newseed_generating_seed">Generazione del seed in corso</string>
+  <string name="newseed_seed_saved_checkbox">Ho fatto il backup di questa lista di parole.</string>
+  <string name="newseed_words_sole_recourse_checkbox">Ho capito che questa lista è la chiave dei miei fondi.\nSono l\'unico possessore.</string>
+  <string name="newseed_next">Avanti</string>
+
+  <!-- //////////////// Pin dialog //////////////////// -->
+
+  <string name="pindialog_title_default">Per favore inserisci il PIN</string>
+  <string name="pindialog_pwd_placeholder">●●●●●●</string>
```
dpad85: I don't think this field should be translated.

PR #31, opened 2020-03-05
Update docker build

araspitzu:
Due to the recent ubuntu releases the updates for ubuntu 18.10 have been dropped and when the docker build invoked `apt-get update` it failed with:
```
E: The repository 'http://archive.ubuntu.com/ubuntu cosmic Release' does not have a Release file.
```
The fix is to bump the ubuntu version to a more recent 19.10.

However this is not enough because some recent changes in the build (perhaps [this](https://github.com/ACINQ/phoenix/commit/c9e49693a4d8bc0ab1a5d19aef76bf16d42f06d3)?) made necessary to explicitly set the version of the ndk used, otherwise the gradle build failed with:
```
android.ndkVersion from module build.gradle is not set
ndk.dir in local.properties is not set
ANDROID_NDK_HOME environment variable is not set
sdkFolder is /usr/local/android-sdk
Because no explicit NDK was requested, the default version '20.0.5594570' for this Android Gradle Plugin will be used
Considering /usr/local/android-sdk/ndk-bundle in SDK ndk-bundle folder
Considering /usr/local/android-sdk/ndk/21.0.6113669 in SDK ndk folder
Rejected /usr/local/android-sdk/ndk-bundle in SDK ndk-bundle folder because that location has no source.properties
Rejected /usr/local/android-sdk/ndk/21.0.6113669 in SDK ndk folder because that NDK had version 21.0.6113669 which didn't match the requested version 20.0.5594570
```
The proposed fix is to set the ndk version in the gradle build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.10
+FROM ubuntu:19.10
 
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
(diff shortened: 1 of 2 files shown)


dpad85:
The change in `versionCode` just increments the build version of the app.

However defining a specific version for the NDK in the gradle build config does make sense, because it will force the development version to be built with the same version as the release (and vice-versa).

PR #38, opened 2020-04-01
Update strings.xml

bitcoinuser:


bitcoinuser:
Remember that if we click on lnurl link, open the same window. So I think Phoenix should create another string removing bitcoin word from the (Scan a QR code containing a payment request or bitcoin address).

PR #39, opened 2020-04-01
Update strings.xml

bitcoinuser:
Do not make sense ask if the user did the backup if the only answer in this case is no.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -300,7 +300,7 @@
 
   <string name="inappnotif_set_up_pin">Your wallet is not protected.</string>
   <string name="inappnotif_set_up_pin_action">Set a PIN</string>
-  <string name="inappnotif_mnemonics_never_seen">Thanks for using Phoenix. Have you made a backup of your wallet?</string>
+  <string name="inappnotif_mnemonics_never_seen">Thanks for using Phoenix. It is very important to back up your wallet so you don't risk losing your bitcoins.</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -300,7 +300,7 @@
 
   <string name="inappnotif_set_up_pin">Your wallet is not protected.</string>
   <string name="inappnotif_set_up_pin_action">Set a PIN</string>
-  <string name="inappnotif_mnemonics_never_seen">Thanks for using Phoenix. Have you made a backup of your wallet?</string>
+  <string name="inappnotif_mnemonics_never_seen">Thanks for using Phoenix. It is very important to back up your wallet so you don't risk losing your bitcoins.</string>
```
bitcoinuser, line 303:
```suggestion
  <string name="inappnotif_mnemonics_never_seen">Hi, it's very important to backup your wallet so you don't risk losing your bitcoins.</string>
```

dpad85: Thanks for the suggestion. We use a question because the user may have restored the wallet (cf #27), or because the user has already backed up the seed, but not checked the confirmation box.

Also why do you want to remove the "Thanks for using Phoenix"? I think it's nice to have a welcome message.

bitcoinuser: Ha ok, thanks for the explanation. The idea to remove the welcome message is because my suggestion was a long phrase. 
But I think when issue https://github.com/ACINQ/phoenix/issues/27 be fixed, the message should be changed.

dpad85:
Superseded by #82

PR #43, opened 2020-04-08
Add dependencies as submodules

Giszmo:


dpad85:
Hi, and thanks for this PR.

I understand that git modules can be handy, but I think using them as a dependency management tool is not correct. We use `eclair-core` and the tor proxy as libraries, which gradle/maven handles well.

If the goal is to be able to easily build the application from a single project, and not have to clone/build these 2 dependencies as well, the correct solution is that ACINQ publishes these libraries to Maven Central, like it's done for bitcoin-lib.

Also as a side note, this PR merges the `mainnet` branch into the `master` branch. They have different purposes, one for production and the other for development, and have their own styling and configurations. For example `master` uses testnet coins, and `mainnet` uses real bitcoins. We could use flavored build to handle this better and remove the separate branch, but this is what we have for now.

PR #46, opened 2020-04-08
Show details

bitcoinuser:
I think here you could use just Show details.


dpad85:
Thanks for your PRs @bitcoinuser !

I think that for wording updates it makes sense to group them into a single branch/PR. This will make for a smaller and more readable git log.

PR #48, opened 2020-04-10
Multistage deterministic build

araspitzu:
This PR improves the deterministic build by **not** requiring the user to provide the dependencies for phoenix, instead we build eclair-core in an intermediate container and Tor-onion-library in the final container.

Note: in the first stage of the build we make the eclair-core artifact and we use the same base image as in the eclair Dockerfile (adoptopenjdk-11) because eclair doesn't compile with JDK8 anymore.
Digest hash of the artifacts @ 1dea4d6
```
baccaf7558f11575da9fb7890337f15a838136ab1d396d3385ed315959f4a723  apk/release/phoenix-7-1dea4d6-testnet-arm64-v8a-release.apk
0d4ab5ff44b4bd93d3f1b47a0179d50674129db23193a499b11027871614982a  apk/release/phoenix-7-1dea4d6-testnet-armeabi-v7a-release.apk
f4aeb7488765a8a1c81a83bf5c813112ad140bb4e5e728b39ea3cc2e15142d07  apk/release/phoenix-7-1dea4d6-testnet-universal-release.apk
3af1f598aeff13a9dd084274d60914ee2382dc67364bedfbd1eb6b4cca7d40b7  apk/release/phoenix-7-1dea4d6-testnet-x86_64-release.apk
4da0bc4a8d08ea831b510e3096c2a1ba78d8c4ef75baf08316dd92e898f87e15  apk/release/phoenix-7-1dea4d6-testnet-x86-release.apk
```
cc: @Giszmo

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -47,6 +47,6 @@ which revision you need to use in the deterministic build.
 ### How to build phoenix deterministically
 
 1. Clone the phoenix project from https://github.com/ACINQ/phoenix
-3. Run `docker build -t phoenix_build . --build-arg ECLAIR_TAG=<GIT_TAG_OF_ECLAIR_CORE>` to create the build environment
+3. Run `docker build -t phoenix_build .` to create the build environment
 4. Run `docker run --rm -v $(pwd):/home/ubuntu/phoenix/app/build/outputs -w /home/ubuntu/phoenix phoenix_build ./gradlew assemble`
 5. Built artifacts are in `.apk/release`
--- a/BUILD.md
+++ b/BUILD.md
@@ -48,5 +48,5 @@ which revision you need to use in the deterministic build.
 
 1. Clone the phoenix project from https://github.com/ACINQ/phoenix
 3. Run `docker build -t phoenix_build . --build-arg ECLAIR_TAG=<GIT_TAG_OF_ECLAIR_CORE>` to create the build environment
-4. Run `docker run --rm -v $(pwd):/home/ubuntu/phoenix/app/build -w /home/ubuntu/phoenix phoenix_build ./gradlew assemble`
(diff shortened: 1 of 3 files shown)


BUILD.md
```diff
@@ -35,18 +35,18 @@ You can check that the corresponding `.jar` file is present in your local maven
...
 2. docker-ce installed
-3. Eclair-core published in your local maven repo, check out the instructions to build it.
 
 ### How to build phoenix deterministically
 
 1. Clone the phoenix project from https://github.com/ACINQ/phoenix
-3. Run `docker build -t phoenix_build .` to create the build environment
-4. Run `docker run --rm -v $HOME/.m2:/root/.m2 -v $(pwd):/home/ubuntu/phoenix/app/build -w /home/ubuntu/phoenix phoenix_build ./gradlew assemble`
-5. Built artifacts are in $(pwd)/outputs/apk/release
+3. Run `docker build -t phoenix_build . --build-arg ECLAIR_TAG=<GIT_TAG_OF_ECLAIR_CORE>` to create the build environment
```
Giszmo, line 50: Where does `<GIT_TAG_OF_ECLAIR_CORE>` come from? How about you merge Could you merge #43 instead?

araspitzu: I'm not a big fan of git submodules but i reckon in this scenario they could prove useful because we can pinpoint the tag of eclair-core from the phoenix repo, however we usually think of eclair-core as a separate library (is used in 1+ projects) and we'd like to keep it that way. Since every release of Phoenix is build against a particular tag of eclair-core `GIT_TAG_OF_ECLAIR_CORE` is the tag that you need to pass to the build, the actual tag should be published in the release notes. 

Note that with this new multilstage build completely removes the need to build external dependencies manually.

PR #51, opened 2020-04-11
fix: Spanish wording, usted -> tú, gender neutral user treatment, som…

mautematico:
…e general string fixes


dpad85:
LGTM! Thanks @mautematico

PR #53, opened 2020-04-14
More options for SSL over Tor for Electrum

t-bast:
If using Tor on port 500001, disable SSL.
Otherwise require a valid certificate.


t-bast:
Using the port is fragile. This will instead be done by expressing whether ssl or plain tcp is used in the url.

PR #54, opened 2020-04-14
Electrum: disable SSL when using Tor

t-bast:
(edited) This PR adds a checkbox to enable/disable SSL for custom Electrum servers when using Tor.

Non-onion Electrum servers must still have a valid certificate.


dpad85:
I ended up hijacking your PR Bastien, and added the checkbox to control SSL when using Tor.

PR #55, opened 2020-04-14
Electrum: use protocol://host:port URIs for custom servers

sstone:
URI can be either tcp://host:port or ssl://host:port for now

### DIFF
--- a/app/src/main/java/fr/acinq/phoenix/utils/Wallet.kt
+++ b/app/src/main/java/fr/acinq/phoenix/utils/Wallet.kt
@@ -32,6 +32,7 @@ import okhttp3.OkHttpClient
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import java.io.File
+import java.lang.IllegalArgumentException
 import java.util.*
 
 object Wallet {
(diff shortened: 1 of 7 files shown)


app/src/main/java/fr/acinq/phoenix/utils/Wallet.kt
```diff
@@ -133,17 +134,24 @@ object Wallet {
...
     // electrum config
-    val electrumServer = Prefs.getElectrumServer(context)
-    if (!Strings.isNullOrEmpty(electrumServer)) {
+    val electrumServerAndProtocol = Prefs.getElectrumServer(context)
+    if (!Strings.isNullOrEmpty(electrumServerAndProtocol)) {
       try {
+        val ssl = when {
+          electrumServerAndProtocol.startsWith("ssl://") -> "strict" // custom server certificate must be valid
+          electrumServerAndProtocol.startsWith("tcp://") -> "off"
+          else -> throw IllegalArgumentException("invalid electrum uri")
```
t-bast, line 143: I think you can't throw here, otherwise all existing users that have electrum servers that didn't specify the protocol will hit that exception...maybe default to `ssl://`? Or do a kind of "migration" for existing users?

skyrdev, line 143: It seems a bit harsh to throw an exception here if the uri does not have a protocol. I'm pretty sure that most users now using this feature have not entered a protocol, so this change would break their app (with a startup failure).

Could we fallback to ssl? It would be the same behaviour as what we have now.

Also this means that the app would support connecting to servers with self signed certificates, right?

dpad85: Also, should not we support `tcp` only if the uri is a `.onion`? Otherwise, the app would support connecting to electrum servers with a self signed certificate.

dpad85:
A better solution is - if the user is connecting to an `.onion` address - to add a checkbox in the Electrum dialog to force the certificate verification. By default, it is unchecked.

If the user is not connecting to an `.onion` address, no checkbox, the app always checks the certificate.

PR #68, opened 2020-05-08
change eclair_version to snapshot

alevchuk:
https://github.com/ACINQ/phoenix/blob/master/BUILD.md refers to instructions to build eclair as a snapshot, so then building phoenix fails because eclair is not found.

I'm not sure how the releases are build, yet that would need be a adjusted slightly... Here I'm fixing the common case.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -133,7 +133,7 @@ dependencies {
 
   // eclair core
   def libsecp256k1_version = "1.3"
-  def eclair_version = "0.3.8-android-phoenix"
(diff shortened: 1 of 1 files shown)


dpad85:
Hi and thanks for this PR. Upgrading the version to `0.3.9-android-phoenix-SNAPSHOT` it's not enough, because there are a few changes in the library (`PaymentInitiator.SendPaymentRequest` and `Eclair.close`).

Commit 2829df7 fixes this, so this PR can be closed.

PR #84, opened 2020-06-23
Change Receber to Resgatar on Portuguese Brazilian

bitcoinuser:


dpad85:
Fixed by 1c12311, thanks!

PR #92, opened 2020-08-01
differentiate lnurl-auth keys per subdomain.

fiatjaf:
Currently Phoenix will cause the same key to be used in filemarket.bigsun.xyz and lnurl.bigsun.xyz, which should be different domain names. Also test.etleneum.com and etleneum.com. Fortunately these are the only examples of this happening in the real world today, so there's no harm in changing this now before it is too late.


dpad85:
Thanks for this PR. I've also fixed the display which will now show the `host` and not the top domain.

dpad85 (approved):

PR #122, opened 2020-12-07
Add and translate new strings to Portuguese Brazilian

bitcoinuser:


dpad85:
Thanks @bitcoinuser Unfortunately I won't be able to add it to the release.

bitcoinuser:
> Thanks @bitcoinuser Unfortunately I won't be able to add it to the release.

I didn't understand. What is the reason?

dpad85:
It's just that the release containing the feature is already made, so your translations will only be added in the release that one.

bitcoinuser:
> It's just that the release containing the feature is already made, so your translations will only be added in the release after that one.

Understood. Thanks

dpad85 (approved):

PR #144, opened 2021-04-26
Enable static remote key feature

pm47:


pm47:
Tested ad6e3cf0, it works well with a mix of channels with/without the feature.

dpad85 (approved):

PR #147, opened 2021-05-01
Keysend prototype

pm47:


pm47:
Rebased on branch [static-remote-key](https://github.com/ACINQ/phoenix/blob/enable-static-remote-key).

Jayware149 (approved):

PR #149, opened 2021-05-06
Add support for LNURL pay

dpad85:
This PR adds support for the LNURL-pay protocol (see https://github.com/fiatjaf/lnurl-rfc/blob/master/lnurl-pay.md), which is a way to offer static invoices. Note that this does not mean that Phoenix can receive payments with LNURL-pay links, only that it can pay them.

This includes support for the comment field and the success actions, which are stored in the payment meta database.


dpad85:
Successfully tested with basic success action.

PR #218, opened 2021-11-05
Adding a Currency Converter tool

robbiehanson:
Helpful when you need to make a payment in a different currency.

We should be able to integrate this tool into the send & receive screens somehow too.


robbiehanson:
Added refresh information & options. If the rates have been refreshed within the last 5 minutes, it says "rates last updated: just now".

dpad85 (approved):
LGTM

robbiehanson:
I needed to force-push because one of the commits wasn't signed. It seems this happens with SourceTree when you cherry-pick or rebase. Even if the original commit was signed, and SourceTree is configured to sign commits, the cherry-pick won't get signed...

Issue #217, opened 2021-10-30
Cannot pay LNURL pay with Phoenix IOS


johnongit:
it seems Phoenix iOS cannot pay an LNURL pay created by  lnbits web app.
Phoenix Android (and other wallet) can pay it.

We can reproduce the bug with the following LNURL p.
https://lnbits.openchain.fr/lnurlp/44

Phoenix returns "Invalid response"

![image](https://user-images.githubusercontent.com/16493486/139536773-1a959a9e-1da4-4388-9e56-3d86f9f5bb4a.png)

robbiehanson:
Confirmed bug. Response from lnbits is:
```json
{"pr":"lnbc1...","routes":[],"successAction":null}
```

Crash occurs because code expects `successAction` to be a JsonObject, and not JsonNull. Fix incoming...

robbiehanson:
Thanks for reporting this bug. It should be fixed in the next release: v1.2.1

PR #219, opened 2021-11-09
Fix for issue #217

robbiehanson:
The LNURL response was:
```json
{"pr":"lnbc1...","routes":[],"successAction":null}
```

This line threw an unexpected exception:
```kotlin
val obj = json["successAction"]?.jsonObject ?: return null
```
Because `jsonObject` request throws `IllegalArgumentException` if it encounters `JsonNull`. Fix:
```
val obj = try {
    json["successAction"]?.jsonObject // throws on Non-JsonObject (e.g. JsonNull)
} catch (t: Throwable) { null } ?: return null
```

I also improved the quality of the error messages while debugging this. So instead of "please try again", it should say something like: "malformed: successAction.message: bad length", which should help fellow developers who are testing with Phoenix.


robbiehanson:
Fixes issue #217

dpad85 (approved):

PR #225, opened 2021-11-26
Translate untranslated Arabic strings

EZ0010:


dpad85 (approved):
Thanks for this update 👍

PR #233, opened 2022-01-06
Import localized recovery phrases

robbiehanson:
This PR allows the recovery phrase to be localized:

french autocomplete

When entering the words, we follow the BIP39 spec. Thus, if using French, typing in "ape" will suggest "apéritif" for autocomplete.

You can select the proper language, as needed:
 
language menu

So the phone's language doesn't have to match the mnemonic language. For example, your phone can be in French, and you can enter an English mnemonic. Or any other combination, such as having your phone in Spanish, and restoring a Czech recovery phrase from the cloud.

Thus, we also show the mnemonic language when displaying the mnemonic:

language indicator

# Major code changes

* There is a new `MnemonicLanguage.kt` file which encapsulates the languages, and various BIP39 related rules.
* The RestoreWallet & Initialization flows now require you to pass the desired `MnemonicLanguage`
* In iOS, we now store a `RecoveryPhrase` struct (with `mnemonics` & `lang`), as opposed to just the `mnemonics`. And all associated functions that previously returned `mnemonics` now return a `RecoveryPhrase`.

# Technical details

Prior to this PR, here's how we stored the seed on iOS:

* we serialized the mnemonics using UTF-8
* we generated a random key (known as the locking key)
* we encrypted the mnemonics (UTF-8 data) using ChaCha20-Poly1305 via the lockingKey
* We stored the resulting `{nonce, ciphertext, tag}` in a file called "security.json"
* We stored the lockingKey in the iOS keychain

With this PR, we follow the same steps. Except instead of storing just the mnemonics (as UTF-8 data), we now store a serialized JSON string, which includes the mnemonics & language code.

The code can handle either one. (An old "security.json" with just mnemonics, or the new version with JSON.) Detection takes advantage of the fact that the JSON string always starts with '{'.


phoenix-ios/phoenix-ios/views/onboarding/InitializationView.swift
```diff
@@ -142,8 +142,10 @@ struct InitializationView: MVIView {
 		let swiftEntropy = AppSecurity.shared.generateEntropy()
 		let kotlinEntropy = swiftEntropy.toKotlinByteArray()
 		
-		let intent = Initialization.IntentGenerateWallet(entropy: kotlinEntropy)
-		mvi.intent(intent)
+		mvi.intent(Initialization.IntentGenerateWallet(
+			entropy: kotlinEntropy,
+			language: MnemonicLanguage.defaultCase
```
dpad85, line 147: This `defaultCase` depends on the user's preferences within the OS, and so it maybe end up being Czech, English, French, or Spanish.

However it seems to me that users nowadays would expect having an English seed by default. I've never had any localized seed for example, but I may not be a good example. What I mean is that some people may be confused if they don't have the regular English words in their seed. 

Also, I don't know if other wallets support non-English seeds? For example, on my current version of Electrum using French as the GUI language, I cannot create a French seed. This would be an issue if people want to recover on-chain funds after a force-close (as mentioned in https://phoenix.acinq.co/faq#my-channels-got-force-closed-how-do-i-recover-my-money).

What do you think about this?

In any case I think it would be a good idea to let people choose what language they use before creating the seed.  The seed initialization screen would need an additional step where people pick a language.

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/utils/MnemonicLanguage.kt
```diff
@@ -0,0 +1,83 @@
...
+        override fun wordlist(): List<String> {
+            return MnemonicCode.englishWordlist
+        }
+        override fun matchMap(): Map<String, String> {
+            return mapOf()
+        }
+    },
+    Spanish {
+        override val code: String get() = "es"
+        override fun wordlist(): List<String> {
```
dpad85, line 17: To be consistent those word lists should be added to bitcoin-kmp in https://github.com/ACINQ/bitcoin-kmp/blob/aa5dc68e25700df53f92f4c31d7f1c0be42fcf8c/src/commonMain/kotlin/fr/acinq/bitcoin/MnemonicCode.kt#L22

dpad85 (commented):
This is a really nice feature but I have concerns about compatibility with other wallets.

Also, when restoring the seed, it is not obvious that the flag in the top right corner applies to the language of the seed (since this is a bit of a niche feature). The UI should somehow tell that this applies to the words of the seed, and that the words entered by the user and that flag must match. 

Maybe move that flag above the words input, and make it an obvious button that says `[flag] Language of my seed` ?

robbiehanson:
> I have concerns about compatibility with other wallets

I will do a deep dive on this and see what I can find.

dpad85:
Just seen this warning in the [BIP39 specs](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#wordlists) (Thanks @pm47 for the link).

> Since the vast majority of BIP39 wallets supports only the English wordlist, it is strongly discouraged to use non-English wordlists for generating the mnemonic sentences.

So it's unfortunate, but I think we should not support localized mnemonics.

robbiehanson:
I understand the motivation for wallet compatibility, but I have to say that the logic seems circular.



That is:

- wallet A doesn't want to support it because wallets B & C don't
- wallet B doesn't want to support it because wallets A & C don't
- wallet C doesn't want to support it because wallets A & B don't



If we continue following this circular logic, then wallets will never support other languages besides English. Which seems in contrast to the BIP39 "Motivation" section:



> A mnemonic code or sentence is superior for human interaction [...]. The sentence could be [...] spoken over the telephone.



But this is only true if the user speaks English - and only [13%](https://www.reference.com/world-view/percentage-world-speaks-english-859e211be5634567) of the world speaks English. If "Bitcoin is for everyone", we need to move beyond English. And if the community wants to "bank the unbanked", my guess is that less than 13% of the unbanked speaks English. Countries like Venezuela & El Salvador have some of the [lowest](https://www.worldatlas.com/articles/countries-that-are-most-and-least-proficient-in-english.html) scores in English Proficiency.



So I have a modest proposal. We have 2 goals:



- maintain wallet compatibility (by following standards)
- bank the unbanked



I think we can achieve both of these goals via:



* If the user taps "create new wallet", we create an English mnemmonic by default (highest compatibility)
* If they press-and-hold "create new wallet", we offer the ability to use another language (with a notice that it may be less compatible with other wallets)
* We offer support for importing a mnemonic in other languages besides English



Thus, on the surface, not much changes. But it's a small nod in the direction of progress. If wallet A offers support, maybe wallets B & C will be encouraged too.

dpad85:
👍 That's a good point, someone has to move first.

I think the long press option will probably not be noticed (unless that's a standard on iOS?). For now I don't have a better option that would not add too many steps in the wallet creation for the default case as well, but is that really important?

If we accept the idea of not creating the seed as soon as the "create new wallet" button is pressed (which is nice UX-wise), but instead agree to have an intermediate screen for some creation options, then we could simply let the user know that the seed is created in English by default. If he changes that to Spanish or French, then we add a notice about compatibility.

robbiehanson:
> when restoring the seed, it is not obvious that the flag in the top right corner applies to the language of the seed (since this is a bit of a niche feature). The UI should somehow tell that this applies to the words of the seed, and that the words entered by the user and that flag must match.

That's a good point. Pushed a change:

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/utils/MnemonicLanguage.kt
robbiehanson: We can do that. Do we only want to add the word lists to bitcoin-kmp? Or do you think it would be helpful if we added the entire `MnemonicLanguage.kt`?

robbiehanson:
Latest iteration:

screenshot 1


screenshot 2

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/utils/MnemonicLanguage.kt
dpad85: Let's keep the additional word lists here for now.

phoenix-ios/phoenix-ios/views/onboarding/InitializationView.swift
dpad85: Mitigated with https://github.com/ACINQ/phoenix/pull/233#issuecomment-1010167040

dpad85 (approved):
LGTM

robbiehanson:
The PR, as it exists now:



- The "wallet creation options" button is hidden
- The options can only be accessed via a long-press, and there is no indication that a long-press option exists. Thus the option is hidden.
- The default is still English.
- The UI supports importing a recovery phrase in multiple languages.
- The UI supports exporting the zprv, which can be used to import the wallet via Electrum.



It's a good first step. We might feel comfortable taking more steps after implementing various on-chain features.

GBKS:
Great to see this being implemented. Would it be possible to detect the default language of the operating system and offer a localized word list only if the default is not English?

Agree with the point about someone having to implement this first and get some momentum going. I am looking into whether it's a good idea to include information and best practices for this in the [Manual backup page](https://bitcoin.design/guide/daily-spending-wallet/backup-and-recovery/manual-backup/) of the Bitcoin Design Guide ([issue](https://github.com/BitcoinDesign/Guide/pull/753)).

robbiehanson:
> Would it be possible to detect the default language of the operating system and offer a localized word list only if the default is not English?

Yes, that's definitely possible. The problem is that, as of right now, many (most?) wallets don't supporting importing a non-english recovery phrase. And thus everyone is afraid to take this step.

For example, suppose Phoenix does this, and we automatically generate a Spanish mnemonic for a user. But then the user discovers they can't import the phrase into Bluewallet, or others. Probably they will blame Phoenix, and tell everyone that we aren't a compatible wallet, or don't follow standards.

That's why we think the first step is for wallets to support **importing** non-english mnemonics. And the Bitcoin Design Guide could be instrumental in this regard.

After a significant number of prominent wallets support importing localized mnemonics, wallet developers will feel more comfortable turning on this feature: defaulting to generating non-english mnemonics for non-english speakers.

dpad85:
The main issue we have with this feature is that, currently, we rely on a third party solution to recover funds on-chain after a force-close, because Phoenix only does Lightning. Any seed we generate needs to be compatible with at least one on-chain wallet for that reason. Our recommended solution today is Electrum, but I don't think they plan supporting non-English seeds.

As mentioned Robbie, having another reputable wallet that would support importing localized seeds would be very helpful. 

Another solution would be to support on-chain recovery within Phoenix, but that would be a lot of work on our side.

GitTankMusic:
BRD closed down, I have 12 BRD wallet Chinese recovery phrase word from BRD, I don't know how can I use the Chinese recovery phrase word to get my coins back, can I use this new feature to recovery my coins form the BRD 12-Chinese recovery phrase word?

Bosch-0:
> But then the user discovers they can't import the phrase into Bluewallet, or others.

Aren't Phoenix recovery phrases only usable within Phoenix currently (as they restore the SCB from the ACINQ node)? I don't see an issue with generating non-English recovery phrases for Phoenix if this is the case. Down the line when when the recovery phrase from Phoenix is interoperable than it would be an issue (by then hopefully other wallets support localized recovery phrases). 

> And the Bitcoin Design Guide could be instrumental in this regard.

Yeah we will get this content in the guide and begin reaching out to wallets to get important non-English recovery phrases as a first step (@GBKS) has already started this. 

Would also be great to have more languages supported, BIP39 is  limited to 10 atm. 

Also flags shouldn't be used to represent languages as many languages can be spoken in one country. From https://www.nngroup.com/articles/flag-problems/

> The problem with using flags as symbols for language choice is that some languages are spoken in many countries. For example, using an American flag Flag of the USA for the English language understandably makes the British upset (they invented the thing, after all) and also irritates Canadians and many others. Of course, using a Canadian flag wouldn't work since many Canadians speak French.

They recommend using a flag to match the geographical location but I'd just not use a flag at all and just have the language name in plaintext (in their respective language).

dpad85:
> Aren't Phoenix recovery phrases only usable within Phoenix currently (as they restore the SCB from the ACINQ node)? I don't see an issue with generating non-English recovery phrases for Phoenix if this is the case. Down the line when when the recovery phrase from Phoenix is interoperable than it would be an issue (by then hopefully other wallets support localized recovery phrases).

Yes that's correct. The issue is when there's a conflict in the channel and funds must go on-chain. That scenario should be rare but it must still be accounted for. In that case, funds must be recovered with an on-chain wallet compatible with seeds generated by Phoenix.

> BRD closed down, I have 12 BRD wallet Chinese recovery phrase word from BRD, I don't know how can I use the Chinese recovery phrase word to get my coins back, can I use this new feature to recovery my coins form the BRD 12-Chinese recovery phrase word?

No you cannot, Phoenix only recovers Lightning channels created in cooperation with the ACINQ node. You can try your luck with 3rd party tools like https://github.com/iancoleman/bip39 (not affiliated with us, be very careful always) to get the private key and import that into Electrum.

GBKS:
The BlueWallet crew just merged a [PR](https://github.com/BlueWallet/BlueWallet/pull/4667) to add the 9 other language wordlists to their import feature. A great step forward that should also help you be more comfortable in adding import support. What are some other wallets we could talk to about supporting this?

robbiehanson:
> What are some other wallets we could talk to about supporting this?

The more, the merrier ! :santa:

robbiehanson:
The branch has been rebased on master. The new focus is on **importing** localized recovery phrases.

Pankyexchange (approved):

robbiehanson:
This branch has been rebased on master. The following changes are present:

It now allows **IMPORTING** a localized recovery phrase:



You can choose the language:



And auto-complete is language appropriate:



The language is represented when you display the recovery phrase within the app:



Since not all wallets allow importing a recovery phrase outside of English, we also allow the user to copy the zprv/zpub. Which means the wallet can be imported into Electrum, for example.



Absolutely **nothing** has changed about the Initialization screen. If the user taps "Create new wallet" then an English recovery phrase will be generated.



If you build-and-go in **DEBUG** mode, then you'll have this additional option in Settings:



Which allows you to select the language (for testing / debugging):

dpad85:
Thanks for this update! this is is a nice change, less ambitious than it initially was since we focus on just importing the seed but it's a good first step.

Also, with Phoenix v2, since we now display funds from closed channels in Phoenix (in the final wallet screen), we're a bit less dependent on other wallets.

> Since not all wallets allow importing a recovery phrase outside of English, we also allow the user to copy the zprv/zpub. Which means the wallet can be imported into Electrum, for example.

This is a good idea. However the method `PhoenixBusiness.walletManager_finalOnChainWallet_xprv` that gets the xpriv should be done in `lightning-kmp` (to be consistent with the xpub). The UI should also warn against storing the file to disk or sharing it on the internet. It could be a footgun. Maybe this `Copy` option should be hidden unless the user does some advanced action.

Another thing is that the words list (in Czech, French and Spanish) should be added to [bitcoin-kmp](https://github.com/ACINQ/bitcoin-kmp/blob/b807d2aefa61b62d598c6769c0668383335e025c/src/commonMain/kotlin/fr/acinq/bitcoin/MnemonicCode.kt#L23) alongside the English words list.

robbiehanson:
> I'm wondering if it was technically possible to convert a non-English recovery phrase into English

Good question. Unfortunately that's not possible. If you take a look at [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed), it details how to go from mnemonic to seed:

> To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic" + passphrase (again in UTF-8 NFKD) used as the salt. The iteration count is set to 2048 and HMAC-SHA512 is used as the pseudo-random function. The length of the derived key is 512 bits (= 64 bytes).

In other words, the process is:
```
pbkdf2-hmac-sha512(input: mnemonic_string, salt: "mnemonic", iterations: 2048) => seed_512_bits
```

So a different mnemonic string (for example, in a different language) would generate a different seed.

phoenix-ios/phoenix-ios/views/configuration/privacy and security/recovery phrase/RecoveryPhraseView.swift
```diff
@@ -854,6 +874,42 @@ fileprivate struct RecoveryPhraseReveal: View {
...
+			.confirmationDialog("What would you like to copy?",
+				isPresented: $showCopyOptions,
+				titleVisibility: .automatic
+			) {
+				// Note: confirmationDialog strips all formatting from Text items.
+				// So we don't get to play with fonts or colors here.
+				Button("Recovery phrase (12 words)") {
+					copyRecoveryPhrase()
+				}
+				Button("Account extended private key (\(xprv))") {
```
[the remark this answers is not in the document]
dpad85: In fact, the xprv export button should be removed. Instead of having 1 secret (the 12-words recovery phrase), Phoenix would now have 2 secrets. It's fine for technical people, but confusing/dangerous for regular users.

This button is only useful to recover funds after a force-close, since we rely on Electrum for that. Which is not good anyway. The correct solution is being able to empty the final wallet directly from Phoenix.

So let's remove this xprv button and merge this PR. We'll then make another PR that support spending from the final wallet. Some work will be needed in lightning-kmp, but that can be done in the next weeks.

dpad85 (changes requested):

phoenix-ios/phoenix-ios/views/configuration/privacy and security/recovery phrase/RecoveryPhraseView.swift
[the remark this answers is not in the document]
robbiehanson: Done in e43191e

phoenix-ios/phoenix-ios/views/configuration/privacy and security/recovery phrase/RecoveryPhraseView.swift
```diff
@@ -854,6 +874,42 @@ fileprivate struct RecoveryPhraseReveal: View {
...
+					copyExtPrivKey()
+				}
+				Button("Account extended public key (\(xpub))") {
```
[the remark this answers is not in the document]
robbiehanson: Done in e43191e

dpad85 (approved):
🚀

PR #236, opened 2022-01-16
(iOS) The Arabic translation

EZ0010:
Hi, after translating the legacy android app I thought of translating the iOS app too.

I built the app to see if it will be RTL, and it became RTL without making any changes to the code, here are some screenshots:
image

There are few issues with RTL (such as the arrows at onboarding) I will make an issue for these.
Also I can get some volunteers to review the translation if needed.


robbiehanson:
Thank you !

Issue #235, opened 2022-01-11
Invoice expiration settings/configuration (Bolt #11)


braydonf:
When attempting a pull-payment with Phoenix from [BTCPayServer](https://github.com/btcpayserver/btcpayserver), as in the case of a refund, it will warn "The BOLT11 invoice must have an expiry date of at least 30 days from submission (Provided was only 6)".

Is there a way to configure the expiration date? If not I think this would make a good feature.

braydonf:
Please see associated issue https://github.com/btcpayserver/btcpayserver/issues/3281

dpad85:
For now this expiry is 1 week and is not configurable (ios or android). That could be useful indeed, we could add that setting next to the default invoice description.

robbiehanson:
Configuration option added to iOS via PR #239

PR #239, opened 2022-01-20
Adding ability to configure invoice expiration length

robbiehanson:
Fixes issue #235 

Screen Shot 2022-01-20 at 17 24 57

### DIFF
--- a/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/controllers/payments/Receive.kt
+++ b/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/controllers/payments/Receive.kt
@@ -20,7 +20,7 @@ object Receive {
         data class Ask(
             val amount: MilliSatoshi?,
             val desc: String?,
-            val expiryDays: Long = 7
+            val expirySeconds: Long = 3600 * 24 * 7 // 7 days
         ) : Intent()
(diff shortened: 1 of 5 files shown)


phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/controllers/payments/Receive.kt
```diff
@@ -17,7 +17,11 @@ object Receive {
     }
 
     sealed class Intent : MVI.Intent() {
-        data class Ask(val amount: MilliSatoshi?, val desc: String?) : Intent()
+        data class Ask(
+            val amount: MilliSatoshi?,
+            val desc: String?,
+            val expiryDays: Long = 7
```
dpad85, line 23: Nitpick: even though this expiry can only be at least 7 days, I'd still make this expiry field a duration in seconds so that the UI has more flexibility.

robbiehanson: Yup, I agree. That's more flexible.

robbiehanson: Done in cc1aaf7

dpad85 (approved):

Issue #191, opened 2021-10-01
Add setting to modify maximum fees for Lightning payments


dpad85:
Phoenix defines a range of fees that the trampoline node is allowed to use for outgoing LN payments. The max fee is currently 12 sat + 0.5%. In some cases this is not enough, and prevent users from paying some nodes.

The iOS app should let the user override the maximum value of those fee. This would be a global setting affecting all outgoing LN payments. Not great but still helpful.

PR #241, opened 2022-01-24
Allows user to configure the maximum fee

robbiehanson:
Fixes issue #191

This PR requires [13c7143](https://github.com/ACINQ/lightning-kmp/commit/13c71436daa1fc8f72166244c4aa143e958c3544) from https://github.com/ACINQ/lightning-kmp/pull/317

Screen Shot 2022-01-24 at 11 49 50

## Technical notes:

My theory is that users will think of this feature a bit like an Ebay auction. With Ebay, if I make a bid of X, I know that I'll only pay Y, where Y is the minimum necessary amount to win the bid. Similarly, if a user sets the max fee to X, they're expecting to pay Y, where Y is close to the minimum fee necessary to make the payment succeed.

So the lightning-kmp library needs a schedule of fees. And our default schedule is a slow ramp up to a conservative max fee. My fear is that, if we simply append the max fee of X at the end of the list, then the user will be paying the max fee everytime the required fee exceeds the conservative maximum.

So I'm thinking that, if our default schedule is something like `[Q, R, S, T, U]`, and the user sets the max to X, then the ideal schedule is more like `[Q, R, S, T, U, V, W, X]`, where `[V, W]` ramp up to X. So there's a bit of clever logic in `TrampolineFees.kt` to accomplish this, along with associated unit tests.

### DIFF
--- a/phoenix-shared/src/commonTest/kotlin/fr/acinq/phoenix/utils/TrampolineFeesTests.kt
+++ b/phoenix-shared/src/commonTest/kotlin/fr/acinq/phoenix/utils/TrampolineFeesTests.kt
@@ -1,0 +1,102 @@
+package fr.acinq.phoenix.utils
+
+import fr.acinq.lightning.CltvExpiryDelta
+import fr.acinq.lightning.TrampolineFees
+import fr.acinq.lightning.utils.sat
+import fr.acinq.phoenix.controllers.payments.MaxFees
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class TrampolineFeesTests {
+
+    @Test
+    fun createTrampolineFees() {
+        val defaultFees = listOf(
+            TrampolineFees(0.sat, 0, CltvExpiryDelta(576)),
+            TrampolineFees(1.sat, 100, CltvExpiryDelta(576)),
+            TrampolineFees(3.sat, 100, CltvExpiryDelta(576)),
+            TrampolineFees(5.sat, 500, CltvExpiryDelta(576)),
+            TrampolineFees(5.sat, 1000, CltvExpiryDelta(576)),
+            TrampolineFees(5.sat, 1200, CltvExpiryDelta(576))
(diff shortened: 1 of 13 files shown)


phoenix-shared/src/commonTest/kotlin/fr/acinq/phoenix/utils/TrampolineFeesTests.kt
```diff
@@ -0,0 +1,102 @@
...
+                ),
+                significantFeeBaseDiff = 90.sat, // 90 >= 90
+                significantFeeProportionalDiff = 1000 // 1000 !>= 900
+            )
+            assertEquals(trampolineFees, defaultFees + listOf(
+                TrampolineFees(35.sat, 1500, CltvExpiryDelta(576)),
+                TrampolineFees(65.sat, 1800, CltvExpiryDelta(576)),
+                TrampolineFees(95.sat, 2100, CltvExpiryDelta(576))
+            ))
+        }
```
dpad85, line 100: We could also check behaviour when the proposed base fee is over the significant diff threshold, but proportional fee is below the default max fee (so default list is truncated). And vice-versa.

dpad85 (commented):
Looks good. I would add a sanity check somewhere. Proportional fee should not be exceed 100% for example. When that happens, an error is shown.

robbiehanson:
Good idea. I'm also thinking an example might be helpful. (Math is hard.) Kinda like we did with the tip summary on the outgoing payment screen.

robbiehanson:
UI has been improved. First, I added a quick explanation:

explanation

Then added an example. Makes the math & fees easier to grok. Note that the example is interactive - user can use plus/minus buttons to see how fees affect payments of different sizes.

example

Added sanity checks to ensure baseFee & proportionalFee are within min/max limits. (This is configurable. I'm not really sure what exact values we want to enforce here. Let me know what you think.)

aboveMax

And added checks as recommended concerning odd mismatches:

badBase

dpad85 (approved):
LGTM. The example section is nice 👍

PR #244, opened 2022-02-01
Fixed Czech translation

iBobik:
"odeslat" is imperative mood, but there should be past - "odesláno"


dpad85 (approved):
Thanks!

iBobik:
Btw, should not the comment been changed also in another translations?

dpad85:
Yes the comment should be fixed in other files as well, but AFAICT this comment is auto-generated by XCode. Ideally the key (`SENT` in this case) should be self-describing, making the comment redundant.

PR #246, opened 2022-02-09
Adding ability to delete failed payments

robbiehanson:
Screen Shot 2022-02-09 at 16 06 07

Screen Shot 2022-02-09 at 16 06 12

Only affects failed payments. Successful payments don't display the "Delete" option.


dpad85 (approved):
👍

PR #247, opened 2022-02-11
Integrated currency converter

robbiehanson:
Imagine the user has selected their preferred currencies:
- bitcoin unit: satoshis
- fiat currency: USD

And then we offer them this selection of currencies:

That doesn't really make sense. And it's probably confusing to the user: "What is mbtc ???"

We already know their preferred currencies. And the currency converter gives us a hint as to the other currencies they're probably dealing with. For example, here's mine:

So this PR updates the list:

And if they select "other", we pop them over to the currency converter. When they're done, they return to the original screen, and we use the value they typed in.


robbiehanson:
For example: Asking for 1,000 MXN:

dpad85:
It looks great!

Small bug I noticed: my preferred bitcoin unit is satoshi, in the invoice editing form, the input is in satoshi. I click on other, the default amount is 100 000 000 sat. Not clicking on anything, I go back to Receive: the input now display 1 btc, it does not use the satoshi unit anymore.

robbiehanson:
@dpad85 Good catch. Bug should be fixed, and I think the integration makes a little more sense now.

Issue #200, opened 2021-10-11
Auto read clipboard in send screen


Bosch-0:
When opening the receive screen if the user has a Lightning invoice copied in their clipboard it should auto read it and give an option to click a button opening that invoice. 

Mockup below:

![image](https://user-images.githubusercontent.com/55287964/136814785-5bdc7f40-0f7d-4c89-a7f3-9997fadddd20.png)

Bosch-0:
Here is a quick mockup integrating your suggestions:

- Icon on top right lets you pick an image. 
- Only show the payment amount and who it is being sent to. 
    - If they have a node alias set you could put this there or if they use an invoice name similar to what blixt does. Showing invoice data probably isn't required here. If they want to they have send see it within the next send details screen. 
    - If no name could just say 'to unknown.' 
- If clipboard is clear have option to manually enter, for Lnaddress for example.
    - May need to add a 'clear clipboard' button to the right of the clipboard popup which would switch to the manually enter screen. This would be if the user has something on their clipboard but wants to manually pay instead. 
- Also added other clipboard states for LNURL-Pay and LNURL-Withdraw on the left.

![image](https://user-images.githubusercontent.com/55287964/137432345-edbde267-afc1-43ec-8217-bffd059cf360.png)

robbiehanson:
I started to implement this. But in testing I noticed something I hadn't considered before:

<img height="450" src="https://user-images.githubusercontent.com/304604/152034171-1fb54cff-2670-4fe2-90fe-6a368235010e.PNG">

On iOS, we can quietly ask **if** the clipboard has a string (yes or no). But if we actually **read** the string, then iOS presents this pop-up. I personally find it to be a little distracting.

(Note that it's quite common to have something in the clipboard - usually a string. So this pop-up will be occurring most of the time we visit this screen.)

Also, we probably shouldn't fetch a LNURL unless instructed to do so by the user. There are security considerations.

So, what we can discover (without performing a URL request):
- Bolt 11 invoice
- LNUrl.Auth
- LNUrl (generic URL)

Open questions for discussion:
- Does the iOS pop-up pose a problem ?
- Will the user care, or is this to be expected ?
- How will this feature be affected by Bolt 12 offers ?

dpad85:
> Does the iOS pop-up pose a problem ?

If it does then the whole auto-read feature has to be dropped. By the way, recent versions of Android show the same warning.

[...]

PR #260, opened 2022-03-31
Auto read clipboard in send screen

robbiehanson:
Fixes issue #200 

We automatically read and parse the clipboard. Without fetching the LNURL, we can detect:

- Bolt 11 invoice
- LNUrl.Auth
- LNUrl (generic URL)

So we append this information to the "Paste from clipboard" button.

I also took this opportunity to improve the UI, based on the mock-ups from issue #200 

clipboard_inspect

And here's how we surface the other types:

inspect_auth

inspect_lnurl


dpad85:
As mentioned by [@Bosch-0](https://github.com/ACINQ/phoenix/issues/200#issuecomment-1085373855)

> One visual improvement suggestion would be the hamburger menu, I think it's protruding a little too much into the main screen region.

We could use the `chevron-up` and `chevron-down` icons from https://feathericons.com/ to have something like this 
image

robbiehanson:
Updated UI:

issue_200_v2

dpad85 (approved):
LGTM 👍

Issue #208, opened 2021-10-15
Add 'Hide sensitive values' feature


Bosch-0:
Having the option to hide sensitive values (balance, transaction amounts etc.) is something quite common in Bitcoin wallets and is a good security / privacy practice to enact in general. There is several ways to do this though I believe the least intrusive is tapping and holding the balance to hide / reveal sensitive information. Design mockup below: 

![image](https://user-images.githubusercontent.com/55287964/137436994-c2ff6b06-73f7-4a9c-bc43-07efabb7e47a.png)

robbiehanson:
Thank you once again for all the help & suggestions.

Bosch-0:
No worries! I love Phoenix, I'd be interested in helping out whenever possible. Just tag me on here or dm me over on the bitcoin desgin Slack - [bitcoin.design ](https://bitcoin.design/)

PR #262, opened 2022-04-07
Option to hide amounts on home screen

robbiehanson:
Fixes issue #208 

- easy toggle on/off option from home screen
- option only affects the home screen, in order to protect the user from prying eyes when opening the app
- option is persistent across app launches

issue_208


robbiehanson:
Updated UI:

issue_208_v2

dpad85 (approved):

PR #271, opened 2022-05-10
Store closing transactions as parts

dpad85:
⚠️  This Pull Request depends on https://github.com/ACINQ/lightning-kmp/pull/333

Closing transactions are now stored in a separated table, in the payments database. This lets us separate the closing transactions information from the `status` field of on-chain outgoing payments, that used to be strongly linked to channel closing.

Several methods in `PaymentsDb` and `OutgoingQueries` have also been renamed, either because the names were too long and unwieldy, or non consistent with the incoming payments queries.

Closing transactions data that are stored in the `status` field of on-chain outgoing payments are read by the `OutgoingStatusData.getClosingPartsFromV0Status` method to ensure backward compatibility.

The `CloudData` wrapper objects also supports closing transactions as parts.


phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/db/cloud/OutgoingType.kt
```diff
@@ -20,15 +19,17 @@ data class OutgoingPaymentWrapper @OptIn(ExperimentalSerializationApi::class) co
     val recipient: ByteArray,
     val details: DetailsWrapper,
     val parts: List<OutgoingPartWrapper>,
+    val closingTxsParts: List<OutgoingClosingTxPartWrapper> = emptyList(),
```
dpad85, line 22: @robbiehanson I would like to have your input on this change.

We have several options to store closing tx parts in this wrapper while ensuring backward compatibility:

**Option 1** : `parts` becomes `part: List<PartWrapper>`, where:
```kotlin
@SerialName("lightning")
data class LightningPart(...): PartWrapper
// json => { "type": "lightning", "id": 123... }

@SerialName("closingtx")
data class ClosingTxWrapper(...): PartWrapper
// json => { "type": "closingtx", "id": 123... }
```
It looks great but causes backward compatibility issues. What to do with legacy data which don't have typed parts? The deserializer does not handle that well.

Solutions could then be :
- add a version to `OutgoingPaymentWrapper` (like `CloudData` is) and do some special rules for data without a version (i.e. those data contain untyped parts);
- or find some trick to let the serializer handle the "untyped" legacy data parts as `LightningPart`s by default.


**Option 2** : we store lightning parts and closing txs parts separately. It is backward compatible, does not add much code, but it is quite ugly.

So I've chosen option 2 for now, because it's the easiest but it may not be the best long term. Wdyt?

dpad85: Thanks! I'll keep it that way then.

dpad85:
> I found an issue...

Fixed with [6bf7d75](https://github.com/ACINQ/phoenix/pull/271/commits/6bf7d758e2c88a211f1f25238f7a1d5d0b88eeb0) and https://github.com/ACINQ/lightning-kmp/commit/fec461846cf6c6f6c6eea9663ef48ec3c6c8b58a

dpad85:
I think the iOS part is done now, so this PR is ready to be merged ? The android UI will be updated afterwards (we made changes to the payment details screen in a parallel PR).

robbiehanson:
> I think the iOS part is done now, so this PR is ready to be merged ?

Yes. The 1 reported issue was fixed, and the iOS UI has been updated.

robbiehanson (approved):

Issue #182, opened 2021-09-21
Can't connet to electrum server with self signed certificate (TCP connect: TLS: bad certificate format)


Emzy:
It seems to me I'm not able to connect to my Electrum server (electrum.emzy.de:50002) because I use a self signed certificate.
I have a self signed because of security considerations. I think there are many electrum servers with that setting.

This is a log from my iPhone but the same problem effects the android version:

```   INFO: 2021-09-21T13:45:56.306215047 | f.a.p.c.config.AppElectrumConfigurationController: intent=UpdateElectrumServer(address=electrum.emzy.de:50002)
   INFO: 2021-09-21T13:45:56.309039115 | f.a.p.managers.AppConnectionsDaemon: electrum server config changed to=Custom(server=ServerAddress(host=electrum.emzy.de, port=50002, tls=SAFE)), reconnecting...
  DEBUG: 2021-09-21T13:45:56.314365863 | f.a.p.managers.AppConnectionsDaemon: electrumControlFlow = TrafficControl(networkIsAvailable=true, disconnectCount=1)
   INFO: 2021-09-21T13:45:56.314957141 | f.a.p.c.config.AppElectrumConfigurationController: model=Model(configuration=fr.acinq.phoenix.data.ElectrumConfig.Random@82d81538, currentServer=ServerAddres (truncated)
   INFO: 2021-09-21T13:45:56.315551042 | f.a.p.c.config.AppElectrumConfigurationController: model=Model(configuration=Custom(server=ServerAddress(host=electrum.emzy.de, port=50002, tls=SAFE)), curre (truncated)
  DEBUG: 2021-09-21T13:45:56.339218139 | f.a.p.managers.AppConnectionsDaemon: disconnecting from electrum
WARNING: 2021-09-21T13:45:56.398197174 | f.a.l.b.electrum.ElectrumClient: TCP receive: Operation canceled
WARNING: 2021-09-21T13:45:56.398231983 | f.a.l.b.electrum.ElectrumClient: closing TCP socket.
   INFO: 2021-09-21T13:45:56.399014949 | f.a.l.b.electrum.ElectrumClient: connection state changed: CLOSED
   INFO: 2021-09-21T13:45:56.400051116 | f.a.p.c.config.AppElectrumConfigurationController: model=Model(configuration=Custom(server=ServerAddress(host=electrum.emzy.de, port=50002, tls=SAFE)), curre (truncated)
[...]

PR #272, opened 2022-05-11
Allow self-signed certificates for custom Electrum server

robbiehanson:
Fixes issue #182 

Notes: 
- This fixes the iOS build (matches latest lightning-kmp changes)
- This PR contains a **breaking change for Android**. In particular, in `ElectrumConfiguration.kt`:

```kotlin
// OLD:
sealed class Intent : MVI.Intent() {
  data class UpdateElectrumServer(val address: String?) : Intent()
}

// NEW
sealed class Intent : MVI.Intent() {
  data class UpdateElectrumServer(val server: ServerAddress?) : Intent()
}
```

So the difference here is that previously we passed a string, and the ElectrumConfigurationController handled parsing it to a ServerAddress. But now the client is expected to directly pass the ServerAddress. Which is because the `ServerAddress` class now allows you to pin a public key.

The solution for now (until Android implements custom UI for pubKey pinning) is to copy the old parsing code from `ElectrumConfigurationController`, which was something like this:

```kotlin
intent.address.matches("""(.*):(\d*)""".toRegex()) -> {
    intent.address.split(":").let {
        val host = it.first()
        val port = it.last()
        val serverAddress = ServerAddress(host, port.toInt(), TcpSocket.TLS.TRUSTED_CERTIFICATES)
     }
}
```

- This PR increases security by pinning the public keys for the default list of Electrum servers. (See `electrumConfs.kt`) To verify these pub key values, you can:

```
$ openssl s_client -connect testnet.qtornado.com:51002
``` 

^ This command will print the server's certificate to the console. (The part that starts with `-----BEGIN CERTIFICATE-----` and ends with `-----END CERTIFICATE-----`) Write that to a file (e.g. `testnet.qtornado.com.pem`). And then use this command to extract the public key from the certificate:

```
openssl x509 -in testnet.qtornado.com.pem -noout -pubkey
```

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/settings/ElectrumView.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/settings/ElectrumView.kt
@@ -33,6 +33,7 @@ import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.window.DialogProperties
+import fr.acinq.lightning.io.TcpSocket
 import fr.acinq.lightning.utils.Connection
 import fr.acinq.lightning.utils.ServerAddress
 import fr.acinq.phoenix.android.*
@@ -45,6 +46,7 @@ import fr.acinq.phoenix.android.utils.mutedBgColor
 import fr.acinq.phoenix.controllers.config.ElectrumConfiguration
 import fr.acinq.phoenix.data.ElectrumConfig
 import kotlinx.coroutines.launch
+import java.net.URI
 import java.text.NumberFormat
 
 @Composable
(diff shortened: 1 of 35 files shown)


dpad85:
I found an issue when testing this on testnet. Using address `testnet1.electrum.acinq.co` with port `51002`, I got the following message : `Server certificate is trusted` after clicking on `Save`, but no info about the certificate. And then I'm stuck because the `Save` button does not do anything, and the `Cancel` button just closes the sheet without saving my custom server.

robbiehanson:
Good catch! Fixed in [01f5078](https://github.com/ACINQ/phoenix/pull/272/commits/01f507852aa087cd9225c5d44a9a2850249bf5b1)

dpad85:
A few nitpicks: the icon used to cycle through the certificate fields is a bit strange (it's usually an icon used to reply). Maybe we could use arrows like this:
image
image


Also, if the certificate is invalid, it might be nice to also print an error message below the `Connecting to xxxx`, in the electrum server screen.

robbiehanson:
> the icon used to cycle through the certificate fields is a bit strange (it's usually an icon used to reply).

Yeah, that's a good point. Changed it to this one:

Screen Shot 2022-05-13 at 09 46 43

robbiehanson:
> Also, if the certificate is invalid, it might be nice to also print an error message below the Connecting to xxxx, in the electrum server screen.

Added notice:

Screen Shot 2022-05-13 at 10 10 55

And we have a NoticeBox on the HomeView, since the user might not think to check the settings:

164541124-6fd9282f-8c07-45e3-80c8-300e27c8f27e

PR #281, opened 2022-05-24
Define fee values

nfillion21:
I do this PR to discuss about where we can define this kind of constant :

Screenshot 2022-05-24 at 16 23 57

I guess these values should be the same on iOS and Android.
I could create a kotlin object in AppConfiguration file.

Screenshot 2022-05-24 at 16 28 44
@robbiehanson What do you think?


robbiehanson:
Yes, that's a good idea. Those values should be shared.

> I could create a kotlin object in AppConfiguration file.

That sounds good to me :+1:

Issue #245, opened 2022-02-07
Inconsistent balance in home screen


dpad85:
On iOS, there's an inconsistency where the spendable balance of a channel is rounded down in the channels screen, but rounded up in the home screen: 

![image](https://user-images.githubusercontent.com/5765435/152768514-c7c53d86-576d-480a-85a6-bba38cfd495b.png)

This channel actually has a balance of 790 millisatoshis. Since we don't display millisatoshis, I think the balance in the home screen should be rounded down.

robbiehanson:
This was a well-timed issue. I was thinking about this the other day, and how we might go about fixing it.



I think part of the problem is that the current codebase is inflexible. It only supports 2 policies:

- Show millisatoshis
- Hide millisatoshis

And if msats are hidden, then it rounds to the nearest satoshi amount - unless that amount is zero, in which case it rounds up to one.

But I think the app needs the flexibility to support additional policies. That is, the above rounding policy probably makes sense for the transaction history table. But maybe not for the wallet balance.

So I've pushed a commit that moves the codebase in the direction of supporting additional policies. As a side benefit, I think it makes the code more readable:

```swift
// before:
Utils.format(currencyPrefs, msat: msat, hideMsats: false)
// after
Utils.format(currencyPrefs, msat: msat, policy: .showMsats)
```

And now we can add new policies:

```swift
Utils.format(currencyPrefs, msat: msat, policy: .showMsatsIfZeroSats)
```

So for the primary balance, we only show msats if:
 `0 < msats < 1_000`

<img height="450" alt="Screen Shot 2022-02-15 at 15 52 50" src="https://user-images.githubusercontent.com/304604/154129352-c3d23982-11f4-4e70-ab31-038e40d03a27.png">

---

I'm not sure if this fully fixes the issue. What do you think? There's also a question regarding the balances within the channels-configuration screen. Should we display millisatoshi's there? Or is there some additional logic we use, for example:

- if any channel has a local or remote balance where: `0 < msat < 1_000`
- then show millisatoshis for all channels

dpad85:
> So for the primary balance, we only show msats if:
> `0 < msats < 1_000`

I think that makes a lot of sense. We should use that policy for the balance display in the home screen and for the user's balance in the channels-configuration screen (only for the user's balance, not for the capacity or for the remote's balance).

[...]

PR #285, opened 2022-06-02
Inconsistent balance in Home vs Channels screen

robbiehanson:
This commit addresses issue #245 

> We should use [the .showMsatsIfZeroSats] policy for the balance display in the home screen and for the user's balance in the channels-configuration screen (only for the user's balance, not for the capacity or for the remote's balance).

This commit updates the channels-configuration screen to use the new policy. Note that previously, the channel balance was being exported (from phoenix-shared) in Satoshi's. So I had to change that to MilliSatoshi's... which probably also means we need another commit here for the Android UI.


dpad85 (approved):
LGTM 👍 

(Android UI has been fixed and PR rebased on master)

PR #287, opened 2022-06-08
Updating UI for "Drain wallet" screen

robbiehanson:
This PR attempts to simplify the UI for the user. Here's what the settings screen looked like before:

ss0

The problem is that the "average user" doesn't know what payment channels are. They are akin to UTXO's on L1 - a technical detail the user is shielded from.

What's worse is that, if the average user wants to drain their wallet, they might mistakenly think that the only way to do this is thru the "Danger zone" screen.

With these ideas in mind, we've made minor adjustments to simplify things for the user:

ss1

ss2

And the force-close channels option is still available via the payment channels screen:

ss5

### DIFF
--- a/phoenix-ios/phoenix-ios/fr.lproj/Localizable.strings
+++ b/phoenix-ios/phoenix-ios/fr.lproj/Localizable.strings
@@ -1182,10 +1182,7 @@
 "The bitcoin mempool is congested and fees are very high. The on-chain swap service has been temporarily disabled." = "Le réseau Bitcoin est saturé et les frais sont très élevés. Le service de swap on-chain a été désactivé temporairement.";
 
 /* label text */
-"The closing transaction is in your transactions list on the __main__ screen. And you can view the status of your channels in the __channels list__ screen." = "La transaction de fermeture est visible dans la liste des transaction, dans l’écran __principal__. Vous pouvez voir le statut de vos canaux de paiement dans l’écran __list des canaux__.";
-
-/* label text */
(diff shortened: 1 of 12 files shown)


phoenix-ios/phoenix-ios/fr.lproj/Localizable.strings
```diff
@@ -1182,10 +1182,7 @@
 "The bitcoin mempool is congested and fees are very high. The on-chain swap service has been temporarily disabled." = "Le réseau Bitcoin est saturé et les frais sont très élevés. Le service de swap on-chain a été désactivé temporairement.";
 
 /* label text */
-"The closing transaction is in your transactions list on the __main__ screen. And you can view the status of your channels in the __channels list__ screen." = "La transaction de fermeture est visible dans la liste des transaction, dans l’écran __principal__. Vous pouvez voir le statut de vos canaux de paiement dans l’écran __list des canaux__.";
-
-/* label text */
-"The closing transactions are in your transactions list on the __main__ screen. And you can view the status of your channels in the __channels list__ screen." = "Les transactions de fermeture sont visibles dans la liste des transaction, dans l’écran __principal__. Vous pouvez voir le statut de vos canaux de paiement dans l’écran __liste des canaux__.";
+"The closing transaction is in your transactions list." = "La transaction de fermeture est visible dans la liste des transaction.";
```
dpad85, line 1185:
```suggestion
"The closing transaction is in your transactions list." = "La transaction de fermeture est visible dans la liste des transactions.";
```

dpad85 (commented):
LGTM 👍 just a small typo.

phoenix-ios/phoenix-ios/fr.lproj/Localizable.strings
robbiehanson: Gracias

PR #293, opened 2022-07-21
iPad support

robbiehanson:
This PR adds support for iPad :grinning:

<br/>

The vast majority of the changes are limited to the iOS UI codebase. But there is one change that affects the shared Kotlin codebase:

The original idea was to:
- change the Home screen to only display "recent" payments
- move the full payments history into a dedicated screen (where we could implement additional functionality in the future)

This change has been implemented (details below). But since this affects Android too, we could optionally split this change into a separate PR. Let me know what you think.

---
#### Kotlin Changes

In order to support fetching "recent" payments from the database, 2 new queries have been added to AggregatedQueries.sq:

- listRecentPaymentsOrder => returns in-flight payments, and recent payments completed after the given date
- listOutgoingInFlightPaymentsOrder => returns only in-flight payments

And there is a new `PaymentsPageFetcher` class that handles the logic for you:

```kotlin
fun subscribeToRecent(offset: Int, count: Int, seconds: Int) { /*...*/ }
```

If seconds is 0 (or negative), it will use `listOutgoingInFlightPaymentsOrder`. If seconds is positive, it will use `listRecentPaymentsOrder`, and launch a task that automatically refreshes the list when the oldest payment expires. So payments automatically get removed from the list when they exceed the desired "recent-ness".


robbiehanson:
UI improvements:

Simulator Screen Shot - iPad mini (6th generation) - 2022-07-25 at 11 24 19

dpad85:
Tested on an actual iPad device and it looks great! 

A few nitpicks:
- in the "scan" screen, the buttons section at the bottom of the screen used to paste data or pick an image should have some bottom padding. The OS puts a black line over that section and it hides parts of the buttons.
- the "View logs" screen is very narrow (since it's constrained to fit the side menu) but displays a lot of data. Not sure if that can be done, but it would be nice if the side menu width could expand when using that screen (and that screen only). Also maybe use a smaller font.

dpad85:
It seems that the "full payment history" button below the list of payments in the Home Screen does not always work. Sometimes pressing it does not do anything.

robbiehanson:
> It seems that the "full payment history" button below the list of payments in the Home Screen does not always work.

Should be fixed with https://github.com/ACINQ/phoenix/pull/293/commits/3f22944924ddd2f207b60a8f1f1e24bb54c5ab33

robbiehanson:
> the "View logs" screen is very narrow but displays a lot of data

Yeah, I don't like that screen either. And the TextView chokes when we try to have it display too much data. It's  an issue on both iPad & iPhone, so I've created a new issue for it: https://github.com/ACINQ/phoenix/issues/294

dpad85 (approved):
LGTM 👍

PR #301, opened 2022-10-06
TorSettings: improve connections title

MarnixCroes:
not sure why the `or` is currently there

removing seems better to me

as it'll now be in the GUI:

```
version=xxx
network=x

---
connections
---
xxx

---
circuits
---
xxx
```


dpad85 (changes requested):
`or` stands for `Onion Router`. It should have been at least capitalized, but `Relay` should be preferred according to the tor specs.

dpad85 (approved):

dpad85:
Thanks!

PR #303, opened 2022-10-12
Feature: Close wallet

robbiehanson:
We previously allowed the user to "drain" their wallet (i.e. move all funds to a different bitcoin wallet). With this feature we're making it easier for the user to:

- delete local data (i.e. wipe app state, as if the app had just been installed)
- delete data stored in their iCloud account

This also follows Apple's new **mandatory** [account deletion policy](https://developer.apple.com/support/offering-account-deletion-in-your-app).

---
#### The new "Close wallet" option is available in Settings:

ss1

---
#### The new "Drain wallet" flow is more interactive and user-friendly:

ss2ss3ss4

---
#### The new "Close wallet" flow:

ss5ss6ss7


robbiehanson:
Updated UI design:

---
#### The "Drain wallet" & "Close wallet" options are both available in Settings:

ss1

---
#### The new "Drain wallet" flow is more interactive and user-friendly:

ss2ss3ss4

---
#### The new "Close wallet" flow:

ss5ss6ss7

#### Warnings for the user

ss8

If the user attempts to close their wallet under certain conditions, we display a warning:
- User has elected manual backup of their recovery phrase, and their balance is non-zero
- User has elected iCloud backup of their recovery phrase, but they are deleting their iCloud backup, and their balance is non-zero

The user must check the boxes before the UI allows them to continue.

dpad85:
Looks great!

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/controllers/config/CloseChannelsConfigurationController.kt
```diff
@@ -46,34 +46,36 @@ class AppCloseChannelsConfigurationController(
...
 
     init {
         launch {
             peerManager.getPeer().channelsFlow.collect { channels ->
 
+                val closingChannelIdsCopy = closingChannelIds?.toSet()
+
                 val updatedChannelsList = channels.filter {
-                    closingChannelIds?.let { set ->
+                    closingChannelIdsCopy?.let { set ->
```
dpad85, line 78: This can be simplified to :

```kt
closingChannelIdsCopy?.contains(it.key) ?: true
```

By the way, the `private var closingChannelIds` variable in the controller that is updated/consumed in different coroutines looks a bit fishy. There should be a `FIXME` comment so that it can be fixed in a later PR.

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/controllers/config/CloseChannelsConfigurationController.kt
```diff
@@ -100,13 +105,16 @@ class AppCloseChannelsConfigurationController(
                     val wallet = walletManager.wallet.value!!
                     val address = wallet.onchainAddress(path)
```
dpad85, line 106: We should rather use `Peer.finalAddress` (from lightning-kmp) instead of computing the address ourselves. This would also allow us to remove the `Wallet.onchainAddress` method (and the tests)

@pm47 can you confirm this is the correct address? Btw this field does not seem to be used (nor tested) in lightning-kmp.

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/controllers/config/CloseChannelsConfigurationController.kt
robbiehanson: I've never been happy with the `CloseChannelsConfigurationController`. The whole thing just feels brittle to me.

From a UI perspective, the user is asking to drain his/her wallet. So they would expect that:

- we show them the total balance (in sats) they should receive
- we close every single channel, and broadcast the appropriate transactions
- any deviation from this means an error/warning screen is displayed

But the code is more complicated. First of all, `Peer` requires that we send a COMMAND for each channel that we want to close. But we don't get a response from this command - we have to listen until we see the channel's state transition to CLOSED.

(This is why we were tracking which channels received the command. So the UI could tell when the requested channels had transitioned.)

But the whole process starts with this brittle-looking piece of code:

```kotlin
val channelsToClose = peer.channels.filter {
    when (it) {
      is Normal -> true
      else -> false
    }
}
```

So if any channels had transitioned to Offline, or some other state, then we don't close them. Which makes sense from the point-of-view of the protocol. But not from the point-of-view of the user. If the device goes offline during the process, we shouldn't just skip some channels. And if there are other edge-cases, the user should be alerted about this.

Right now the shared code just kinda does a fire-and-forget of the CMD_CLOSE. But I think there should be a lot more logic here.

robbiehanson: Fixed in a945b4d1

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/controllers/config/CloseChannelsConfigurationController.kt
robbiehanson: Fixed in e2f5762

(It checked, and `Peer.finalAddress` does generate the same address as the previous code. So it does look like this is the proper replacement.)

pm47: > @pm47 can you confirm this is the correct address for force-close? Btw this field does not seem to be used (nor tested) in lightning-kmp.

Yes it's the correct address, you can use it. This are is in-progress though, so the api may change.

dpad85: Great! can you also remove the `Wallet.onchainAddress` method that should now be unused, as well as the relevant tests in `WalletTest`?

robbiehanson: Done in [3576d04](https://github.com/ACINQ/phoenix/pull/303/commits/3576d043b4996fec108eec70d1fec824663e8f27)

PR #305, opened 2022-10-17
Distinguish between "balance is 0" and "balance is unknown"

robbiehanson:
Previously the Home.Model was setup as:
```kotlin
data class Model(
    val balance: MilliSatoshi,
    val incomingBalance: MilliSatoshi?,
    val paymentsCount: Long
)
```

However, this has the unintended consequence that the UI cannot distinguish between "balance is zero" vs "balance is unknown". This was been more noticeable since iOS 16, where the app & UI seem to load very quickly, before the channel states in the database have been read. Perhaps this is because iOS 16 launches apps quicker, or perhaps it's related to how many channels the wallet has. Whatever the case, the consequence is the same: the app quickly launches and proudly tells the user that she has ZERO sats in her wallet :scream:

The fix is simple, but also requires a corresponding change in Android:
```kotlin
data class Model(
    val balance: MilliSatoshi?,
    val incomingBalance: MilliSatoshi?,
    val paymentsCount: Long
) 
```

In iOS I changed the UI to display 3 dashed circles (kinda like loading symbols) while we're waiting for the balance to be fetched.

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/components/ProgressView.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/components/ProgressView.kt
@@ -30,12 +30,14 @@ fun ProgressView(
     text: String,
     modifier: Modifier = Modifier,
     padding: PaddingValues = PaddingValues(16.dp),
-    space: Dp = 8.dp
+    progressCircleSize: Dp = 20.dp,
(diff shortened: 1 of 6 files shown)


dpad85 (approved):
Nice change!
I've updated the Android app, we are using an animated spinner when the balance is unknown.

PR #312, opened 2022-10-25
More recent options

robbiehanson:
New options for the "recent payments" configuration on the Home screen:

Screen Shot 2022-10-25 at 15 35 40

With matching changes on the Home screen:

Screen Shot 2022-10-25 at 15 34 44


dpad85 (approved):
I think Option 2 should be the default option, with 3 or 5 as the default value. Option 1 does not seem very useful to me now that I think about it. Option 3 is a good addition.

Regarding the UI, the screen looks imposing and it feels like we don't really know what's the correct solution and just leave it to the user to decide. That's not really blocking, we can go ahead with this, but we should strive to keep the wallet **_look_** as simple as possible (I don't mean dumbed down with no options, just sleek and easy to grasp).

If you also feel like option 1 is redundant, we can remove it.

Another idea if we want to keep all those options is to hide the sliders for the options that are not selected.

robbiehanson:
> but we should strive to keep the wallet look as simple as possible (I don't mean dumbed down with no options, just sleek and easy to grasp).

Agreed. Let me re-think this UI.

robbiehanson:
Updated UI design:

Screenshot 2023-01-13 at 16 56 29

So now the "recent payment" options match the look of the currency options.
And  we push the picker to another screen (also matching the currency options):

Screenshot 2023-01-13 at 16 56 37

dpad85 (approved):
LGTM 👍 

FYI the android app only displays the last 3 payments. We use the `subscribeToAll` method to feed a payments list that used by both the Home Screen and the Payment History screen, though the Home Screen only takes a [subset of this list](https://github.com/ACINQ/phoenix/blob/aedac7f69ac7764e4185aedfcaeae291dfa56779/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/home/PaymentsViewModel.kt#L70-L76). I'm not sure we will use the the `subscribeToInFlight` or the `subscribeToRecent` methods on Android, so if you ever need to tweak those 2 methods it will not impact android.

PR #313, opened 2022-11-08
Fixes for latest lightning-kmp commits

robbiehanson:
This PR is easiest to review per commit.

#### Commit [b781597](https://github.com/ACINQ/phoenix/commit/b7815974dd9c866624f9793b9a220f35c3cd2017)

Fixes issues from:
- https://github.com/ACINQ/lightning-kmp/pull/391

This PR made several serialization classes internal, and exposed a new generic Serialization API. We were directly using a small number of the (now internal) serialization classes. So they've been copied into this project.

_Note that there's a small bug in this PR causing 1 unit test to fail: `SqliteChannelsDatabaseTest.basic()`. Until that's fixed, you can apply the 1-liner patch from the conversation as a workaround._

#### Commit [c3f67b1](https://github.com/ACINQ/phoenix/commit/c3f67b18cafa1ed673c9f12dd9f558bc76fddc81)

Finally fixed the foreign-key-constraints problem on iOS. I discovered that I forgot to enable the foreign-key-constraints when creating the drivers for the unit test.

#### [8a54bf5](https://github.com/ACINQ/phoenix/commit/8a54bf582d60db0839e41f673134821cea459261)

Fixes issues from:
- https://github.com/ACINQ/lightning-kmp/pull/395

With the new dual-funding flow, there are now separate fees:
- `serviceFee` for the channel creation & supplied liquidity
- `fundingFee` for L1 transaction that funds the channel (i.e. miner fee)


dpad85 (approved):
Looks good. I think we need to add non regression tests that read cloud blobs that were serialised using the old `vX` serialisers from lightning-kmp.

There's already 3 tests doing that in `CloudDataTest`:
- `read_legacy_uncompressed_pubkey`
- `outgoing__parts_before_closing_txs`
- `incoming__receivedWith_newChannel_legacy_no_uuid`

Those 3 tests should be moved to another file (like `NonRegCloudData`) because most tests in `CloudDataTest` are simply roundtrip tests.

This should be done in another PR.

Issue #294, opened 2022-08-08
iOS: View logs doesn't work properly


robbiehanson:
The issue is that, when the logs are long, they don't appear properly in the TextView. It would be preferable if we could open the logs in a 3rd-party application.

robbiehanson:
The bug has been fixed, and the fix will be available in the next (major) app release: v1.6.0

PR #325, opened 2022-12-19
Remove broken log viewer & fix log file export

robbiehanson:
The built-in log viewer within Phoenix iOS doesn't work well. And it never has. It's just a TextField with the contents of the raw log file. If the log file gets too big, the TextField chokes, and doesn't display anything. Further, it doesn't have any of the things one would want in a log viewer:

- ability to search
- displaying error messages in red

So I think it's better to simply remove this, and allow the user to open the log file from any 3rd party app. Which they can do via the existing share functionality.

This fixes issue #294

In addition, I also noticed that `LogsConfigurationController.kt` was exporting the log file immediately. That is, if the user enters that view, the log gets exported. Even if the user never opts to export the file.

I think it's preferable if we export the logs on demand. Further, not exporting the logs on demand may result in missed logging information. For example:

- user generates an invoice, and sends it to a peer
- user navigates to Log screen, and waits
- user receives payment
- user taps "Share the log file", thinking this includes the recent payment

Previously, with the log file generated prematurely, the exported log file would **not** include the recent payment. This PR fixes that problem.

Note: This PR likely requires changes on Android. This is a low priority fix.


dpad85 (approved):
Nice change it's a good idea to remove the "view logs" option from iOS as it does not work well 👍 

Note about logging on Android: `kodein-log` on Android delegates everything to SLF4J which handles logs rolling and writing files to disk as well as branching to Logcat (the default Android logging tool). The Android app actually does not use the logs controllers or `LogMemory`. Those were specifically made for iOS because there are no good logging utilities on iOS as far as I know (maybe that's changed since?).

I think that eventually we should remove this logic from the `phoenix-shared` module (especially the writing to disk part where we rely on `kodein-memory`, which brings unnecessary complexity) and do things natively in Swift (if you think that's worth it).

robbiehanson:
> I think that eventually we should remove this logic from the phoenix-shared module (especially the writing to disk part where we rely on kodein-memory, which brings unnecessary complexity) and do things natively in Swift (if you think that's worth it).

That's a good idea. I didn't know Android wasn't using it. I'll create an issue for it.

robbiehanson:
Let's merge this **after** PR #327

PR #327, opened 2022-12-20
Merge iOS and Android hotfix branches

dpad85:
This PR merges into master the current hotfix branches for the legacy Android app and the iOS app:
- https://github.com/ACINQ/phoenix/tree/android-legacy-hotfix (android 1.4.23, 1.4.24)
- https://github.com/ACINQ/phoenix/tree/ios-hotfix (iOS 1.5.2, 1.5.3, 1.5.4)

Which contain fixes for the legacy trampoline bits and lnurl-auth.


phoenix-ios/phoenix-ios/BusinessManager.swift
```diff
@@ -182,14 +182,16 @@ class BusinessManager {
 		log.trace("loadWallet()")
 		assertMainThread()
 		
+		if (business.walletManager.isLoaded()) {
+			return false
+		}
+
```
dpad85, line 188: @robbiehanson Probably something you want to take a look at, I think it's correct but it's worth checking.

robbiehanson: It's still correct. We have multiple guards setup, but this is a good place to have defensive programming.

robbiehanson:
A file managed to sneak into the repository using 2 different names: "LNUrlManager.kt" & "LnurlManager.kt"

This caused some confusion on my machine, which has a case-insensitive filesystem. So I tried to fix that.

I also added some unit tests to check for deserialization problems concerning the class name changes. Everything looks good.

robbiehanson (approved):

Issue #304, opened 2022-10-15
Coindesk API provides inaccurate fiat conversions rates


zikziko:
While using wallet i came across strange currency conversion. The coindesk api reports vastly different prices from other providers or even from their own web version.
https://api.coindesk.com/v1/bpi/currentprice/CZK.json
https://www.coindesk.com/price/bitcoin/ 
the problem seems to affect all non USD currencies.

The behavior can be huge missmatch for EUR
https://api.coindesk.com/v1/bpi/currentprice/EUR.json
given the recent EUR drop in price against USD. As of writing(19 169 USD) API price:  18 673 EUR   vs "real price":  19 720 EUR.

dpad85:
Are you using Phoenix on Android or on iOS?

FYI, on both the Android app and the iOS app, the EUR/BTC and USD/BTC tickers come from the blockchain.info API, so we do not use Coindesk there, and the rates are accurate (right now the rate in the app is 1 btc = 19 974 EUR, which is correct).

Coindesk is used for the currencies other than USD/EUR, and that's where there can be discrepancies. We're aware of that and we have plans to fix that in the future.

zikziko:
I've seen the EUR only in the API and didn't check phoenix itself, sorry for this misinformation. I'm using the app on Android. I'm glad you are aware of the issue for other currencies. It seems to me that coindesk is using some old exchange rates between USD and other currencies which is causing this behavior.

robbiehanson:
Notes:
- We've had reports of inaccurate SEK, Swedish crowns (ticket 47508568)
- We don't currently support [CNH](https://wise.com/us/blog/cny-vs-cnh), but we would like to

robbiehanson:
Exchange rate for COP is very inaccurate recently. For example, today:
- coindesk: 1 USD = 4,219 COP
- everybody else: 1 USD = 4,757 COP

robbiehanson:
The raw JSON from the coindesk feed contains this disclaimer:

> "This data was produced from the CoinDesk Bitcoin Price Index (USD). Non-USD currency data converted using hourly conversion rate from openexchangerates.org"

However, if I pull directly from the source, I get very different exchange rates:

- direct from CoinDesk API: 1 USD = 4,218.99 COP
- direct from OpenExchangerRtes API: 1 USD = 4,765.08 COP

[...]

Issue #298, opened 2022-09-22
Add Cuban Peso (CUP) price with free market rate.


Ventura94:
Add the price of the Cuban Peso (CUP) with the free market rate, for this they could use the API yadio.io
API de CoinDesk: 1 USD = 25,75 CUP
Yadio API: 1 USD = 185 CUP

robbiehanson:
Historical context: https://github.com/ACINQ/phoenix-kmm/issues/204

PR #331, opened 2023-01-03
CoinDesk exchange rates no longer accurate

robbiehanson:
We've had multiple reports of the CoinDesk exchange rates being inaccurate, most recently in issue #304.

I can attest to this personally, as I've noticed that the exchange rate for COP (Colombian Peso) has been very inaccurate recently. For example, the other day:

- coindesk: 1 USD = 4,219 COP
- actual rate: 1 USD = 4,757 COP

The raw JSON from the coindesk feed contains this disclaimer:

> "This data was produced from the CoinDesk Bitcoin Price Index (USD). Non-USD currency data converted using hourly conversion rate from openexchangerates.org"

However, if I pull directly from the source, I get very different exchange rates:

- direct from CoinDesk API: 1 USD = 4,218.99 COP
- direct from OpenExchangerRates API: 1 USD = 4,765.08 COP

So the OpenExchangeRates API has an accurate value, while CoinDesk does not. So if CoinDesk is indeed pulling from this source, then they're doing so poorly. It may be that they're no longer updating those rates...

Thus I went looking for alternative sources. The trouble is that most of the well-known sources for fiat exchange rates (e.g. OpenExchangeRates.org) are either not free, or their free service requires registration, and your account token is rate-limited (e.g. 1,000 API requests per month).

One possible solution to this would be to aggregate these exchange rates server-side, and provide an API to Phoenix clients. We can ponder the pros and cons of this approach later. For now, I've discovered an alternative solution: the Coinbase API

Coinbase's free & public API supports all the same currencies we were using Coindesk for... except certain countries that the USA is sanctioning:

- CUP: Cuban Peso
- ERN: Eritrean Nakfa
- IRR: Iranian Rial
- KPW: North Korean Won
- SDG: Sudanese Pound
- SOS: Somali Shilling
- SYP: Syrian Pound

So what I've done is replace coindesk with coinbase for all the other fiat currencies. That means 142 currencies that were previously using Coindesk, and getting stale exchange rates, are now using coinbase and getting accurate exchange rates.

For the 7 currencies listed above... well... we're still using coindesk for them :man_shrugging:

Other technical notes:

- I removed the "Maltese Lira", as they adopted the Euro back in January 2008
- I added CNH, which is the Chinese Yuan (offshore market) rate. We had a request for this on Jitbit awhile back.
- I also added CUP_FM, which is the Cuban Peso (free market) rate. From issue #298.

### DIFF
--- a/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/CurrencyManager.kt
+++ b/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/CurrencyManager.kt
@@ -66,14 +66,26 @@ class CurrencyManager(
      */
     private val highLiquidityMarkets = setOf(FiatCurrency.USD, FiatCurrency.EUR)
 
+    private val specialMarkets = setOf(FiatCurrency.ARS_BM)
+
+    private val missingFromCoinbase = setOf(
+        FiatCurrency.CUP, // Cuban Peso
+        FiatCurrency.ERN, // Eritrean Nakfa (accidentally overwritten by ERN altcoin?)
+        FiatCurrency.IRR, // Iranian Rial
+        FiatCurrency.KPW, // North Korean Won
+        FiatCurrency.SDG, // Sudanese Pound
+        FiatCurrency.SOS, // Somali Shilling
+        FiatCurrency.SYP  // Syrian Pound
+    )
+
     /**
-     * We use a number of different API's to fetch all the data we need.
+     * We use a number of different APIs to fetch all the data we need.
      * This interface defines the shared format for each API.
      */
     private interface API {
         /**
          * How often to perform an automatic refresh.
-         * Some API's impose limits, and others simply don't refresh (server-side) as often.
+         * Some APIs impose limits, and others simply don't refresh (server-side) as often.
          */
         val refreshDelay: Duration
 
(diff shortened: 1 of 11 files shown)


phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/CurrencyManager.kt
```diff
@@ -410,78 +489,176 @@ class CurrencyManager(
...
+                        )
+                    }
+                }
+            }
+        } ?: listOf()
+
+        return fetchedRates
+    }
+
+    private suspend fun fetchFromCoinDesk(
```
dpad85, line 660: There's some advanced performance logic in this method that was added because the Coindesk API does not provide a single point of entry to fetch everything in one request and we had to chain ~146 requests sequentially. But that's not the case anymore since now we only use Coindesk for 7 currencies, so maybe can remove this code? Or at least update the comment to show that it's not really a concern anymore.

I'm also wondering if the new query to Coinbase (which fetches a large-ish JSON) could have any impacts on the iOS notification-service extension? If I recall correctly it was quite sensitive to performances issues.

dpad85 (commented):
Looks great! I think the about screen for the iOS app needs to be updated to mention the new sources.

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/CurrencyManager.kt
robbiehanson: > I'm also wondering if the new query [...] could have any impacts on the iOS notification-service extension?

That's a good question. I will investigate.

robbiehanson: > Or at least update the comment to show that it's not really a concern anymore.

Done in 1ccb122. I still prefer to fetch all the Url's first, and write to the database after, since it was overall faster in previous tests.

robbiehanson:
> I'm also wondering if the new query [...] could have any impacts on the iOS notification-service extension?

I investigated this. It doesn't seem to cause problems. The memory limit is 24 MB, and we're using around 9.5 MB, with a high watermark of 14 MB. This seems to be the case whether we pull rates from coinbase or not.

I think the problem we had initially is that:
- the CurrencyManager was updating ALL currencies
- the CoinDesk API was creating 146 separate HTTPS connections

Now we disable auto-refresh in the CurrencyManager, and only refresh the target fiat currency. So that's a max of 2 HTTPS connections. (e.g. for COP, we need to hit coinbase for the COP/USD rate, and blockchain.info for the USD/BTC rate).

(_Note that coinbase does have the USD/BTC and EUR/BTC rate. So we could theoretically drop blockchain.info, and just pull that info from coinbase._)

robbiehanson:
> I think the about screen for the iOS app needs to be updated to mention the new sources.

Done in 9df320e

dpad85 (approved):

PR #333, opened 2023-01-09
Fixing iOS 16 crashes by switching to NavigationStack

robbiehanson:
We have been seeing crashes on iOS 16 that have a stack trace like this:

- 0: SwiftUI: UIKitNavigationBridge.containingNavControllerFromLastAttemptedPop.getter
- 1: SwiftUI: UIKitNavigationBridge.containingNavControllerFromLastAttemptedPop.getter
- 2: SwiftUI: UIKitNavigationBridge.isInTransientPop.getter
- 3: SwiftUI: UIHostingController.resolveRequiredBridges(in:allowedActions:)
- 4: SwiftUI: UIHostingController.didMove(toParent:)
- 5: SwiftUI: objc UIHostingController.didMove(toParent:)
- 6: UIKitCore: -[UIViewController removeChildViewController:notifyDidMove:]
- 7: UIKitCore: -[UINavigationController removeChildViewController:notifyDidMove:]
- 8: UIKitCore: -[UIViewController dealloc]
- 9: UIKitCore: -[UINavigationController viewDidDisappear:]

The crash has affected dozens of different devices, but exclusively those running iOS 16 (all known versions including 16.0, 16.1, 16.2)

I've been able to reproduce the crash on my own devices. (Restore a wallet, and crash occurs as UI transitions from IntroductionView to MainView)

In researching the issue ([1](https://developer.apple.com/forums/thread/716310)), the most reliable solution is to replace all instances of NavigationView with a NavigationStack. But since NavigationStack is iOS 16 only, we had to jump thru some hoops.

(Note: Switching to a NavigationStack also fixes annoying iOS 16 bugs, such as ScrollView content being oddly scrolled down when first displaying a view within a navigation context.)


robbiehanson:
Developer notes:

Apple's official recommendation is to adopt NavigationStack when your minimum deployment target is iOS 16. This is completely unhelpful if `NavigationView` exhibits bugs in iOS 16, and developers are forced to upgrade.

After switching to a NavigationStack, I noticed that `NavigationLink(destination:isActive:)` doesn't work properly. In fact this style of NavigationLink is now deprecated. So with NavigationStack, you have to switch to one of the new NavigationLink's. Of course, the new NavigationLink's are iOS 16 only... So you have to use `if #available(iOS 16, *)`, ideally with a clean wrapper function.

Similarly, `NavigationLink(destination:tag:selection:label:)` doesn't seem to work either, IF using an `EmptyView()` for the label.

Sometimes the bugs were only present on the actual device, and worked fine on the simulator.

There is a `NavigationStack(path:root:)` which takes a `NavigationPath` binding as a parameter. But there are gotchas when using a NavigationPath.

First, the use of `.navigationDestination(isPresented:destination:)` doesn't work properly with a NavigationPath. That is, it will push the view onto the NavigationStack just fine. However, the item isn't represented in the NavigationPath. (You can verify by checking `navPath.count`.) So it basically breaks programmatic navigation, which is the entire point of using a NavigationPath in the first place. This forces you to switch from `.navigationDestination(isPresented:destination:)` to `.navigationDestination(for:destination:)`.

By itself, this isn't a big deal, and isn't that hard to refactor. However, after doing all the work, I encountered an infinite loop bug in SwiftUI. After debugging for hours, I discovered that the root cause seems to be nested `.navigationDestination(for:destination:)`. That is, NavigationStack.rootView has a `.navigationDestination()`. And then adding another `.navigationDestination()` to subviewX will cause SwiftUI to break, and enter an infinite loop.

I found a similar report on [stackoverflow](https://stackoverflow.com/questions/74592169/nested-navigationdestination-infinite-loop-in-swiftui), and a bunch of people complaining about the issue on the [forums](https://developer.apple.com/forums/thread/708592).

The most frustrating part is that I spent an entire day trying to create a small project that reproduces the issue. (So I could better understand the bug, and also report it to Apple.) But I was unsuccessful. This frustration was echoed in the forums, with others trying and failing to reproduce the bug in small sample projects. Some attributed the issue to the use of CoreData, or `@AppStorage`. We're not using either, so my best guess is that it's a complicated bug within the inner workings of SwiftUI.

Whatever the case, I was forced to abandon using NavigationPath. And rely on the same tricks we've been using for programmatic navigation in the past.

Note that even if Apple fixes this in a 16.X revision, you're still going to have users running old 16.Y revisions. Meaning it probably won't be safe to switch to NavigationPath until at least 17.0.

PR #345, opened 2023-03-02
Improved handling of iOS notifications permissions

robbiehanson:
In the past, we used "silent" push notifications. These were **unreliably** delivered to the application. In the event they were actually delivered, they allowed Phoenix to receive a payment as long as:
- Phoenix was running in the background
- "Background app refresh" was enabled for Phoenix

However, we recently switched to the more reliable method of using notification service extensions (#280). Interestingly this doesn't rely on "background app refresh" permissions. Instead it relies on the Notification Settings:

&nbsp;

&nbsp;

In order for Phoenix to receive payments in the background, iOS must launch and run our notification service extension (`UNNotificationServiceExtension`) when it receives a push notification.

- If notifications are disabled, iOS will NOT launch our extension
- If notifications are enabled, but all 3 options are disabled (lock screen + notification center + banners), then iOS will NOT launch our extension
- If ONLY "banners" are enabled, then iOS will NOT launch our extension when the screen is locked

The only reliable configuration is when:
- notifications are enabled
- either "lock screen" or "notification center" is enabled

What's more, if we don't properly request permission for notifications within the app, then permissions are implicitly disabled.

**Traditionally**, asking for permission means a screen like this is displayed to the user:

You're only allowed to ask the user once. Further requests are ignored by iOS - nothing is displayed to the user, and your permissions don't change. Because of this Apple has recommended that you ask the user for permission when it's related to an action they're taking within the app.

That is the **traditional** way to get notification permissions. But there is another way that I think works better for us: **provisional** permission.

If instead you ask for provisional permission to display notifications, iOS will immediately grant your request, and you will be given the following permissions:

(_You'll notice that we're granted permission to run our notification service extension._)

And the first time you display a notification, the user will see something like this:

So to summarize the situation on iOS:
* the only way to reliably receive payments in the background is with a notification service extension
* iOS will only launch your extension if the user allows a minimal subset of notifications
* you can get this minimal subset for free with provisional notifications (it doesn't even prompt the user)
* or you can try to get the full set of permissions by prompting the user and hoping they tap "Allow"
* but you have to decide between the 2 options, because you're only allowed to ask for authorization once (subsequent requests are ignored by iOS)

What to do ... ?

I think the most important thing is actually receiving the payment in the background. If the user receives the payment, but the app icon doesn't have a badge... well that's better than NOT receiving the payment. If they receive the payment, but the notification wasn't on the lock screen, but only in the notification center... well that's also better than NOT receiving the payment.

Apple's decisions have made things difficult for us, and a bit confusing for the users. Take another look at the screenshots from the Notification settings above. It **looks** like you're just controlling where the little push notification window is allowed to be displayed. Except you're actually not - you're turning on & off background payments. Because you're controlling whether or not Phoenix is allowed to run in the background.

So probably the most elegant solution for now is to request provisional permissions. This means we won't have to prompt the user for permission when they first start using the app. And by default our extension has permission to run.

If the user disables "background payments" then we'll warn them:

&npsp;&nbsp;

We'll add a new "Background payments" configuration:

&nbsp;&nbsp;

Tapping on "Customize in iOS settings" takes the user directly to the proper screen. Also the user can remove the amount from the notification:

And if they fully disable background payments, we explain the problem and how to fix it:


robbiehanson:
The root of the problem we're trying to solve is this:



^ This is the screen within iOS Settings that allows the user to control notifications.

If I turn off notifications for Signal, I **still receive the messages** within Signal. I'm only turning off the little push notification windows. I quickly learn that the iOS notification settings simply control what gets displayed in iOS. (E.g. banners, badges, etc).

And it's the same for Facebook. If I disable notifications for Facebook, **I still see the posts in my feed**. I'm only disabling notifications on iOS.

It's the same for WhatsApp, or Twitter, or Yahoo Finance, or any other app that I can think of.

The problem is that it's **not** the same with Phoenix. If the user disables notifications, they **do not receive the payments** within Phoenix. And this is not what they expect.

(_I believe this problem is solved with async payments. But those are still a work in progress. What we have right now is the best we can do at the moment._)

So we have to explain to the user:

* disable notifications ==> disable background payments


> The settings > display options > background payments Screen lists several options that can be checked by the user ; are those options in addition to the iOS settings for Phoenix?

I'm trying to use this screen to explain the above. Tapping  any of those checkboxes takes you to the iOS settings. (Note that it's not possible to programmatically change those settings within Phoenix. The user must make these changes within iOS.) So the effect is something like this:

[Video Demo](https://www.dropbox.com/s/amteunb1mjwsilr/ios-notifications.mp4?dl=0)



I think my goal with this design was simply to link the two things mentally: notification settings <=> background payments



They're related. They're in the same screen. I have to enable the "notification center" to enable background payments. And if I'm worried about privacy: there's an option right there to hide amounts.



I don't know if I accomplished my goal. The screenshots can be confusing. But interacting with the UI is more intuitive, and easier to grok. 



Maybe I should clearly explain on that screen that "show in notification center" is required.

dpad85:
Ok I understand now. I was not able to test on a device so I was just looking at the screenshots and they made me think that you need to fiddle with those settings within Phoenix. But it's just an information screen.

And by the way, we don't have to detail why background payments are enabled. My suggestion is:
- if background payment is enabled:
    - remove the checkboxes;
    - state that background payment is enabled 👍 ;
    - have a button redirecting to the iOS setting screen.
- if background payment is disabled, keep what you already have. The screen is a bit intense but that's not a bad thing. 

> Maybe I should clearly explain on that screen that "show in notification center" is required.

It's already done in the "disabled" scenario so I think you're good.

dpad85 (approved):

Issue #344, opened 2023-02-23
LBP fiat currency addition 


marcondy:
It would be great to have integrated the real market rate of LBP ( Lebanese Lira ) 

[API documentation ](https://yadio.io/api.html) 

https://api.yadio.io/convert/1/BTC/LBP

Thank you in advance.

dpad85:
Good idea and thanks for the API endpoint, it seems there's a large discrepancy with the rate we use. I don't think we'll be able to deliver this quickly though.

marcondy:
> Good idea and thanks for the API endpoint, it seems there's a large discrepancy with the rate we use. I don't think we'll be able to deliver this quickly though.

Thank you for your prompt response, no worries as long as it's going to be correct eventually. 

Majority of converters online use the "official" rate , therefore it's been wrong for over three years now.

Thank you again

robbiehanson:
Is there a "common name" for the free market rate (to differentiate it from the official rate) ?

For example, in Argentina it's commonly called the "blue market" (or "dólar blue" or "mercado azul"). So we named the currency "Argentine Peso (blue market)" - ARSbm

When we added support for Cuban Pesos I wasn't aware of any such "common name", so we simply called it "Cuban Pesos (free market)" - CUPfm

We could do the same here: "Lebanese Pound (free market)" - LBPfm

Unless there's a common name already in circulation? Let me know your thoughts.

marcondy:
> 

Yes , they common name for street USD value is the "Black Market"

robbiehanson:
Support for this feature has been merged into master, and should be available in v1.6.1 (or v2.0.0 whichever arrives first)

PR #348, opened 2023-03-24
Adding support for Lebanese Pound (black market)

robbiehanson:
As requested in issue #344 

We were already fetching this exchange rate, since we were already using Yadio for another currency. So we just needed to add the currency, and a bit of extra logic to pull it from the fetched data.

Naming considerations:

- With the Argentine Peso, the common name for the "unofficial" rate is "blue market"
- With the Cuban Peso, I wasn't aware of a common name, so I chose the term "free market"
- I [asked](https://github.com/ACINQ/phoenix/issues/344#issuecomment-1481387635) about the Lebanese Pound and was told it's "black market"

Future work:

The `CurrencyManager` is currently fetching data from 5 different sources. It's not a ton of data, and the overhead isn't that bad. BUT... it's unnecessary work for nearly every user. For example, if you're only using the Euro or the US Dollar, then you only really need to fetch from 1 source.

The `AppConfigurationManager.preferredFiatCurrencies` already tells us which currencies are needed by the user within the UI. So I have a separate commit I'm working on that will optimize the CurrencyManager such that it only pulls the data it needs to. But it's a less trivial change so it's better to have it in a separate PR.

### DIFF
--- a/phoenix-android/src/main/res/values/strings.xml
+++ b/phoenix-android/src/main/res/values/strings.xml
@@ -703,9 +703,9 @@
 
     <!-- ////////////// currencies ////////////// -->
 
-    <string name="currency_ars_bm">Argentine Peso (Blue Market)</string>
-    <string name="currency_cup_fm">Cuban Peso (Free Market)</string>
-    <string name="currency_lbp_bm">Lebanese Pound (Black Market)</string>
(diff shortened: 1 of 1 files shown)


dpad85 (approved):
LGTM, I updated the android app as well.

Issue #341, opened 2023-02-08
Background app refresh: outdated information


robbiehanson:
Now that we have transitioned to using a notification-service-extension, the information surrounding background app refresh is a little bit outdated.

For example, we say that "Background app refresh" must be enabled to receive payments in the background. This was correct when using "silent push notifications", but is technically incorrect with a notification-service-extension. However, we're currently forcing it to be correct by unregistering for push notifications, based on the "background app refresh" status.

However, the watch tower task DOES require background app refresh. So if the user disables it, we need to warn them somewhere within the app.

PR #351, opened 2023-03-30
Display warning if watchtower is disabled

robbiehanson:
We recently improved the handling of iOS notification permissions (PR #345) because "background app refresh" is no longer a requirement for background payments. However "background app refresh" IS required for the watchtower task that runs every few days.

So this PR adds a warning about it, with information about how to fix it.

&nbsp;&nbsp;

If the user taps the "Fix it" button, they're taken directly to the settings screen in iOS where they can re-enable "background app refresh" for Phoenix.


robbiehanson:
Fixes issue #341

dpad85 (approved):
👍

PR #354, opened 2023-03-31
Show original fiat amount

robbiehanson:
For most people, their unit-of-account is the local fiat currency. Maybe someday in the future it will be in sats, but for now it's usually the local fiat currency.

For example if Alice lives in Madrid, then she thinks in Euros. She has a good idea of what things cost - in Euros. If she travels to Colombia and sees a "lunch special" for 30,000 COP ... she has no idea if that's cheap or expensive until she converts the Pesos to Euros. And it's the same if she's told the lunch special is 22,500 sats.

And when she remembers how much she paid for something, she remembers the value in Euros. "I had an amazing lunch for under 6 Euros !"

The problem is that Phoenix displays the payment history using the **current** fiat value. That is, by taking the original payment amount in Bitcoin, and then converting it to fiat using the **current** fiat/bitcoin exchange rate. This is incredibly frustrating because your memory of how much something cost doesn't match what is displayed in Phoenix:

"🤯 I **know** I paid less than 6 Euros for that meal ! 😡"

However, Phoenix knows what the original fiat amount was. (We've been storing that information since PR #259) 

This PR adds an option to display the original fiat amount in your payment history:

&nbsp;&nbsp;


dpad85:
This is a great change. By default, are you showing the original fiat amount? I think it should be the default.

Not sure about that, but in the payment details screen, we could add a clock icon (or a help `?` icon) next to the fiat amount, to show that it's the amount "then". It could make the screen too busy though.

robbiehanson:
> By default, are you showing the original fiat amount? I think it should be the default.

Yes, it's currently set to be the new default value.

> in the payment details screen, we could add a clock icon (or a help ? icon) next to the fiat amount, to show that it's the amount "then"

That's an interesting idea. I'll experiment with it.

dpad85 (approved):

PR #358, opened 2023-04-18
Fixing TLS negotiation over Tor

dpad85:
This PR fixes an issue with TLS over Tor on iOS.


dpad85:
@robbiehanson I just rebased your commit on `master` to fix the compilation issues caused by the changes on lightning-kmp `master` branch, and also fixed the android module as well. But otherwise it should be good to go.

PR #381, opened 2023-07-17
Recheck for payment confirmation when a new block is mined

robbiehanson:
The `payment.confirmedAt` flag wasn't always being set properly, because this collection:
`paymentsDb.listUnconfirmedTransactions().collect { ... }`

only fired when:
- you restarted the app
- you initiated another unconfirmed tx (list changes)

So if you create a new unconfirmed tx, and then wait, the above collection won't trigger when the tx is finally mined.

The fix is to simply tweak the collection to also fire when there's a new mined block:
```kotlin
combine(
  paymentsDb.listUnconfirmedTransactions(),
  appConfigurationManager.electrumMessages
).collect { }
```

Note that we're not using `electrumClient.notifications` because we'd have to first filter that for instances of `HeaderSubscriptionResponse`. But we also don't want to wait for a new block to start monitoring. What we really want is a `StateFlow<HeaderSubscriptionResponse?>` which AppConfigurationManager already provides.


dpad85 (approved):
Looks good!

PR #395, opened 2023-08-17
Merge channels when upgrading an existing iOS wallet to splice

dpad85:
This PR adds a method that consolidates all existing channels into one new single channel with support for dual funding.

This is especially useful for the iOS app. It allows existing iOS wallets with non-dual funding channels to migrate to dual-funded channels.


robbiehanson:
When the user launches the new version on iOS for the first time, then they're greeted with this screen (if they already have a wallet):



<br/><br/>
I've added some nice little fireworks animations that explode on the screen in random locations.

When the user clicks "get started", then we call the `consolidateChannels` function. And if it finishes successfully, then this view disappears, and the user sees the normal wallet screen. If all went well, this means their channels were all closed, and they have a new channel with zero-conf, meaning their balance is immediately available.

However, there are a lot of edge cases.

* If we call `consolidateChannels` before the Peer is connected, it thinks there aren't any channels that need to be consolidated.
* It's also possible the user has 24 channels, and only half of them have moved from `Offline` to `Normal`
* It's also possible the user has a wallet, but has zero channels.
* Or maybe the Peer is online, but the ElectrumClient is still offline
* Or maybe the user doesn't want to merge channels, they want to drain their wallet and start fresh

For these reasons:

* If the function returns `NoChannelsAvailable` we treat this as success
* If the function returns some other error, we present the error message and 2 buttons: Retry & Abort
* If the user long-presses on the title ("Phoenix v2 has arrived") for 5 seconds, we present 2 buttons: Get started & Skip for now



Now if the user skips, or aborts, or the `consolidateChannels` function otherwise fails to finish the job, then:

* The user has access to the wallet, and can perform operations like "drain wallet" or "force close channels"
* However, access to the "Receive" and "Send" screens are disabled:



<br/><br/>
Note how "Receive" and "Send" are in red. This is triggered whenever the `canConsolidate` function returns true. And if the user taps these buttons, they're prompted to perform the merge again:

pm47 (approved):

PR #397, opened 2023-08-18
Use maven version 3.9.4 in Dockerfile

TKone7:
I could not build the Docker image for building the app since it seems the Maven version 3.9.3 is no longer available. I assumed it's alright to use 3.9.4 instead.


dpad85:
Yes we can use maven 3.9.4. In fact actually already pushed a commit that update the dockerfile, see https://github.com/ACINQ/phoenix/commit/be4f0258240430b0efe1fec7c7625ddec6d60654, so this PR can be closed.

TKone7:
Awesome thanks 🎉

PR #407, opened 2023-08-29
Completed German translation for important_strings

myxmaster:
Native German speaker here, this is my translation for the important_strings.


dpad85 (approved):
Amazing! Thanks a lot for this work. It will be available in version 2.0.5.

PR #410, opened 2023-08-30
German strings.xml

myxmaster:
And here is the strings.xml translation. Thank you for making this great app :)


dpad85 (approved):
LGTM, thanks again!

PR #419, opened 2023-09-17
Update TRANSLATION.md

jooray:
Correct links and fix double reference to the same file


dpad85 (approved):
Good catch, thanks for the fix.

PR #424, opened 2023-09-21
(ios) When scanned QR code is just a website link

robbiehanson:
It's happened to me multiple times. I open up Phoenix, scan a QR code (which I assume is an LNURL), and I get this confusing error message:

It's often the case that the QR code was just a website. And if you open the link in your browser, then the website will help you generate the lightning invoice.

Other times I was just confused. Like when I saw this at a restaurant that said "Pay with crypto. Scan here"

Except that's the link for the menu. But I was very confused when I scanned it with Phoenix...

So this PR makes the experience more helpful for the user:


dpad85 (approved):
Looks good! We'll have to do that on Android as well.

PR #425, opened 2023-09-21
Log the project's git commit and lightning-kmp version

dpad85:
The goal of this PR is to be able to track exactly what commit and what version of `lightning-kmp` Phoenix is using at a given time, to help with debugging. These informations are printed in the logs, when the `NodeParamsManager` starts.

To do this in Eclair, we put the [git-commit in `eclair-core` manifest.](https://github.com/ACINQ/eclair/blob/96ebbfe78918799d809a29f2c7a2f1e57e25d84f/eclair-core/pom.xml#L65-L69)

The solution here is similar: in the gradle build file of `phoenix-shared`, we added a new task `buildVersionsTask` that generates a file in `phoenix-shared/build/generated-src/<package>`:
```kotlin
object BuildVersions {
    const val PHOENIX_COMMIT = "735a8bcfbbc64b0fb08d27ba46449c0b9aef3e91"
    const val LIGHTNING_KMP_VERSION = "1.5.6"
}
```
This task is started every time the sources are compiled (and its output is marked as a source directory for Android Studio) with:
```gradle
kotlin.srcDir(task.map { it.destinationDir })`
```

We could use the same technique in `lightning-kmp` (and all our other kmp libraries like bitcoin-kmp).


dpad85:
@t-bast let me know what you think about this, and if we should do the same for `lightning-kmp`, `bitcoin-kmp`, `secp256k1-kmp`.

PR #426, opened 2023-09-22
(android) Show recently rejected payments in Home

dpad85:
This PR adds a message that summarises recently rejected payments (due to the user's fee policy). 

On Android we already show a system notification when such events occurred, but these notifications are easy to miss, (system gives them low priority, or system notifications are disabled). The user is confused about why the payment seemingly never arrived - so we need to display a message.

This message only counts offchain payments rejected less than 3 hours ago. Clicking on that message redirects to the notifications list.

@robbiehanson this is similar to what was done in https://github.com/ACINQ/phoenix/pull/423 although I preferred not to show the full details of 1 single rejection (it takes a lot of place). It's more discreet that way, and the user can still get the details by clicking on the message.

### 1 payment rejected with 2 important messages

We display two messages, one about the rejected payment, and the other like before, displaying the important with highest priority. Clicking on any of these messages redirects to the notifications list screen.

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/home/HomeNotices.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/home/HomeNotices.kt
@@ -21,6 +21,7 @@ import android.content.Intent
 import android.net.Uri
 import android.os.Build
 import android.provider.Settings
+import android.text.format.DateUtils
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.*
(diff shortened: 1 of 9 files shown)


robbiehanson:
> We display two messages, one about the rejected payment, and the other like before, displaying the important with highest priority.

This is a great improvement! I will copy this on iOS.

> This message only counts offchain payments rejected **less than 3 hours** ago.

Maybe the time duration should be longer. If I miss a payment while sleeping, it would be nice to still see it on the Home screen.

> 1 payment rejected recently

Mild suggestion: "1 incoming payment rejected recently" (possibly more clear, but not convinced this is an improvement)

robbiehanson (approved):

PR #439, opened 2023-10-09
(ios) background receive improvements

robbiehanson:
This PR includes a number of improvements related to issues that have recently affected users:

* There was a bug in the notify-service-extension (notifySrvExt), the background process that handles receiving incoming payments if the Phoenix app isn't running in the foreground. (Remember this runs as a completely separate process.) The notifySrvExt was always using the default liquidity settings (e.g. 5,000 max sats). This is now fixed.
* It's possible that the user launches the main app when the notifySrvExt is running. In the past this would always result in the early termination of the notifySrvExt. However if the notifySrvExt was connected, and potentially processing the incoming payment, this could result in the payment failing. Now the mainApp can yield to the notifySrvExt, and postpone connecting to the lightning peer as needed. (Note: this doesn't affect other connections, such as the connection to the electrum server)
* When the mainApp yields to the notifySrvExt, it displays the status "Receiving in background" within the AppStatusPopover
* The "skip absolute fee check for lightning" toggle button wasn't working correctly. It was always initialized to OFF within the UI, regardless of the actual configuration.
* The "channel management" screen now uses a "Save" button, so the user is made aware that changes take effect after tapping the "< Save" button.


phoenix-ios/phoenix-ios/officers/BusinessManager.swift
```diff
@@ -65,6 +65,12 @@ class BusinessManager {
 	///
 	public let canMergeChannelsForSplicingPublisher = CurrentValueSubject<Bool, Never>(false)
 	
+	/// Reports whether or not the notifySrvExt process is running, and connected to the Peer.
+	/// When this is true, we (the main app process) are prevented from connecting to the Peer
+	/// (otherwise doing so would force-disconnect the notifySrvExt, which may be processing an incoming payment)
+	///
+	public let srvExtConnectedToPeer = CurrentValueSubject<Bool, Never>(false)
```
dpad85, line 72: I'm concerned there will be inevitable edge cases/race conditions where this flag would end up always preventing the peer to connect. Is there a timeout so that we eventually always end up on our feet?

robbiehanson: Yes. The way it works internally is:
* after the notifySrvExt connects to the peer, it stores a timestamp in the shared UserDefaults
* it updates this timestamp every 2 seconds while it's alive & running
* the main app listens to this property, and assumes the notifySrvExt may still be alive if the timestamp is within the last 5 seconds

So we're not dependent on the notifySrvExt completing properly and performing some clean-up operation. It can simply crash, and the main app will recover gracefully.

dpad85 (approved):
🚀

PR #448, opened 2023-10-17
Add timeout to Peer connection

dpad85:
This PR adds support for the new timeouts added in https://github.com/ACINQ/lightning-kmp/pull/541.

These timeouts are dynamic (the Electrum connection timeout has been updated too) :
- if there has been many failed consecutive connection attempts, the timeouts will be longer. Otherwise they are short to make the connections as snappy as possible.
- If Tor is enabled, the timeouts are increased as well.


robbiehanson (approved):
Tested on simulator and device. Connections seem to be working smoothly. Even the flaky Electrum testnet server we have in our list isn't bothering me anymore.

PR #449, opened 2023-10-17
(ios) Improved lightning fee as a percentage

robbiehanson:
There is some user confusion about the lightning fee. The actual fee amount is: 4 sats + 0.4%. However, in the past the lightning fee was always displayed simply as a percentage. So this meant the user might see "0.5%" or "0.7%" when sending smaller payments. They might proceed to think this is the fee for all payments. 

So the new policy is, for larger payments (> 40,000 sat), we simply display the percentage as before:

But for smaller payments, we display "0.4% + 4 sat" to be more precise:

Hopefully this clears up some confusion.


dpad85 (approved):
Note for the future, if the trampoline fee can be overridden, we'll have to adapt that logic.

PR #458, opened 2023-10-26
Import channel

robbiehanson:
I added a shared implementation to phoenix-shared.


dpad85 (approved):
LGTM

I updated the Android app to use the new shared method in `ChannelsImportHelper`.

Issue #442, opened 2023-10-11
Camera stays active


myxmaster:
When you click 'Send', the scanner opens and Android status bar indicates camera usage. When you cancel the scanner or go back, Android camera usage indicator stays.

Phoenix Android 2.0.10
Phone: Samsung Galaxy S20+

dpad85:
It looks like this issue is specific to Samsung, I'll investigate.

kotasjan:
@myxmaster is right. I also noticed this bug on my Google Pixel 7 a few days ago. It is IMHO issue of all Android phones. You just don't see it on devices with Android 11 and lower because this new privacy feature was introduced with Android 12. More info about it [here](https://source.android.com/docs/core/permissions/privacy-indicators).

I looked at the code briefly and I will create a pull request. Unfortunately, I didn't manage to build the app and test it because the setup with all the legacy dependencies is quite complex, and I am still running into some issues with Maven. 🥲

Here is the video showing the bug:

https://github.com/ACINQ/phoenix/assets/17073543/4c004bd8-e1da-4b5a-b42e-bf3ea9ebf474

P.S. The hat is there because I didn't have a tool to blur my transactions properly. 😄

PR #460, opened 2023-10-27
(android) Fix camera keeps running after leaving scanner view (#442)

kotasjan:
Please test the changes. I am quite confident these changes will work, but I didn't have a chance to build and run it due to hurdles with the project build setup. 😕

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/ScanDataView.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/ScanDataView.kt
@@ -21,7 +21,6 @@ import android.Manifest
 import android.content.Intent
 import android.net.*
 import android.provider.*
-import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.shape.RoundedCornerShape
@@ -185,7 +184,7 @@ fun ReadDataView(
                 onScannedText = onScannedText
             )
             CameraPermissionsView {
-                DisposableEffect(key1 = model) {
+                DisposableEffect(key1 = model, key2 = initialInput) {
                     if (model is Scan.Model.Ready && initialInput.isNullOrBlank()) scanView?.resume()
                     onDispose {
                         scanView?.pause()
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/ScanDataView.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/ScanDataView.kt
@@ -180,16 +180,16 @@ fun ReadDataView(
     Box(Modifier.fillMaxSize()) {
 
         if (initialInput.isNullOrBlank()) {
-            BackHandler {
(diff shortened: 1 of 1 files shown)


dpad85:
Thanks, I will look into that ASAP.

phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/ScanDataView.kt
```diff
@@ -179,6 +180,9 @@ fun ReadDataView(
     Box(Modifier.fillMaxSize()) {
 
         if (initialInput.isNullOrBlank()) {
+            BackHandler {
+                scanView?.pause()
+            }
```
dpad85, line 185: This changed can be rolled back if we use a `DisposableEffect`.

phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/ScanDataView.kt
```diff
@@ -216,7 +220,10 @@ fun ReadDataView(
                 modifier = Modifier.fillMaxWidth(),
                 text = stringResource(id = R.string.btn_cancel),
                 icon = R.drawable.ic_arrow_back,
-                onClick = onBackClick
+                onClick = {
+                    scanView?.pause()
+                    onBackClick()
+                }
```
dpad85, line 226: This change can be rolled back if we use a `DisposableEffect`.

dpad85 (changes requested):
I was able to reproduce issue #442 on a Pixel device with Android 14 (though on a Pixel device with Android 12, the camera does not stay active all the time).

I also tested your changes, and pausing the scanner does fix the issue - however it needs to be applied on other cases (e.g. when copy-pasting). A better solution is to simply use a `DisposableEffect` in the `CameraPermissionsView.onPermissionGranted` callback, instead of a `LaunchedEffect`:

```kotlin
DisposableEffect(key1 = model) {
    if (model is Scan.Model.Ready && initialInput.isNullOrBlank()) scanView?.resume()
    onDispose {
        scanView?.pause()
    }
}
```

If you give me push authorisation on your branch, I can update your PR with that change.

kotasjan:
> I was able to reproduce issue #442 on a Pixel device with Android 14 (though on a Pixel device with Android 12, the camera does not stay active all the time).
> 
> I also tested your changes, and pausing the scanner does fix the issue - however it needs to be applied on other cases (e.g. when copy-pasting). A better solution is to simply use a `DisposableEffect` in the `CameraPermissionsView.onPermissionGranted` callback, instead of a `LaunchedEffect`:
> 
> ```kotlin
> DisposableEffect(key1 = model) {
>     if (model is Scan.Model.Ready && initialInput.isNullOrBlank()) scanView?.resume()
>     onDispose {
>         scanView?.pause()
>     }
> }
> ```
> 
> If you give me push authorisation on your branch, I can update your PR with that change.

Thank you for the feedback. You are tight with `DisposableEffect` - it is a much cleaner solution. 👍 I have added you to contributors, so you should be able to push your changes now.

dpad85 (approved):
Thanks again @kotasjan

PR #472, opened 2023-11-17
(ios) Keep-alive integration

robbiehanson:
This integrates the keep-alive mechanism from [lightning-kmp #550](https://github.com/ACINQ/lightning-kmp/pull/550).

* when we receive a `SensitiveTaskEvents.TaskStarted` we call `beingLongLivedTask`
* when we receive a `SensitiveTaskEvents.TaskEnded` we call `endLongLivedTask`

Each call uses a unique identifier that is extracted from the parameters in `TaskStarted`/`TaskEnded`. So if these events are broadcast multiple times, the system can easily ignore them.

In addition, iOS will "punish" our app if we abuse the longLivedTask system. That is, if we always ask for background execution time. Which would happen if the lightning-kmp library didn't always emit a `TaskEnded` in every situation. So to protect against this we've also added a backup call to `endLongLivedTask` that will be executed automatically after a timeout.

This updated architecture is now also used for payments, which are driven by the PeerEvent's: `PaymentProgress`, `PaymentSent`, `PaymentNotSent`.

### DIFF
--- a/phoenix-ios/phoenix-ios/officers/BusinessManager.swift
+++ b/phoenix-ios/phoenix-ios/officers/BusinessManager.swift
@@ -76,7 +76,7 @@ class BusinessManager {
 	private var fcmToken: String? = nil
 	private var peerConnectionState: Lightning_kmpConnection? = nil
 	
-	private var longLivedTask: UIBackgroundTaskIdentifier = .invalid
+	private var longLivedTasks = [String: UIBackgroundTaskIdentifier]()
 	
 	private var paymentsPageFetchers = [String: PaymentsPageFetcher]()
 	private var cancellables = Set<AnyCancellable>()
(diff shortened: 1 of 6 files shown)


phoenix-ios/phoenix-ios/officers/BusinessManager.swift
```diff
@@ -153,14 +153,21 @@ class BusinessManager {
...
+					let paymentId = paymentProgress.request.paymentId.description()
+					self.beginLongLivedTask(id: paymentId)
+					
+				} else if let paymentSent = event as? Lightning_kmpPaymentSent {
+					let paymentId = paymentSent.request.paymentId.description()
+					self.endLongLivedTask(id: paymentId)
+					
+				} else if let paymentNotSent = event as? Lightning_kmpPaymentNotSent {
+					let paymentId = paymentNotSent.request.paymentId.description()
+					self.endLongLivedTask(id: paymentId)
```
dpad85, line 170: If I understand https://github.com/ACINQ/lightning-kmp/pull/550 correctly, `SensitiveTaskEvents` are only published for splicing updates. That's why on Phoenix's side we also need to monitor outgoing payment events. If we did not do that, we would not trigger long lived tasks for when sending LN payments, which would be a regression.

If that's correct, we should add a comment saying that in the long run, `SensitiveTaskEvents` will also cover the LN payment use case and then we'll be able to remove this custom section.

Is that correct @pm47 @robbiehanson ?

dpad85 (commented):
LGTM! Feels great to remove code from the `PaymentManager` 😃

phoenix-ios/phoenix-ios/officers/BusinessManager.swift
pm47: Yes that's correct, it's only a proof of concept for now, it also has the advantage to cover a case that couldn't be handled from the outside of lightning-kmp. We can later extend it to other tasks like payments.

Issue #441, opened 2023-10-11
Improve Screen Lock protection


dpad85:
Phoenix should provide more options for app access control, each with their own trade-offs:
1) biometrics authentication: the recommended option. It enables fine-grained control on a device that is shared between several users (e.g. a family), reasonably secure, but less private and prone to hardware malfunction
2) the device's PIN/password: no fine control and probably less secure, but it's robust
3) a custom password: private & secure, but users can lock themselves out. This can also cause confusion with bip39 passphrases, or with the wallet's recovery phrase.

Currently, 1) is supported on Android and iOS. 2) is supported on Android as a fallback. 3) is not supported.

### Enabling these options

The app access screen should let the user select one or several of the 3 options above. 1) should be prominent, 2) and 3) should be advanced settings.

In all cases, the seed should be backed up before enabling any of these options. Or at least show a visible warning.

### Authenticating

When authenticating, if the user picked several authentication methods, he will first select one, then can authenticate.

### Fallback

In last resort, the user should be able to enter the recovery phrase to open the app. See #339.

robbiehanson:
> In all cases, the seed should be backed up before enabling any of these options. Or at least show a visible warning.

This is "low hanging fruit" and would go a long ways in helping the user. We have had several reports from users who have broken their hardware, and Face ID no longer works. They contact us because they enabled Face ID in Phoenix, and now they're locked out of the app.

These users can be put into 2 groups:
A) those who have a backup of their recovery phrase
B) those who don't

For (A) it's not a problem, as they can simply re-install the app. For (B) it's not so good...

robbiehanson:
Fixed via #478 & #560

Issue #455, opened 2023-10-26
Face-ID necessary, no Code possible 


notrelevant11:
After my TrueDepth Sensor of the iPhone is broken I‘m not able to access the Phoenix app anymore.
If I start the app it will use FaceID with „Please enable Face ID“ olthough I’ve deactivated FaceID systemwide.
There is no automatic use of the Apple Code like on most other apps.

thanks in advance

robbiehanson:
A "passcode fallback" option has been added. It should be available in the next released version of Phoenix.

PR #478, opened 2023-11-28
(ios) App Access Control: Device passcode fallback

robbiehanson:
Partially addresses issue #441

> Phoenix should provide more options for app access control, each with their own trade-offs:
> 
> 1. biometrics authentication: the recommended option. It enables fine-grained control on a device that is shared between several users (e.g. a family), reasonably secure, but less private and prone to hardware malfunction
> 2. the device's PIN/password: no fine control and probably less secure, but it's robust 

Option 2 was missing from iOS, but fixed via this PR.

&nbsp;

If the user:
* enables biometrics (Face ID / Touch ID)
* but does not enable the passcode fallback
* and has not backed up their recovery phrase

then we display this warning:


robbiehanson:
> I think we can just remove that warning altogether : users need to backup their seed regardless of the app-control setting. It's true that a broken biometrics hardware can lock the user out if the passcode fallback was disabled, but that's just an edge case among many and we can't handle all of them.

I agree. And that screen already displays a warning icon next to the "Recovery Phrase", if the user hasn't performed a backup.

Done in 880114a470a5b2793721d744479053cbad5fd15e

robbiehanson:
> Is it possible to let the user only use the Passcode option, without having to enable FaceID/TouchID?

I do not believe this is possible. My understanding is that you have to use [LAContext](https://developer.apple.com/documentation/localauthentication/lacontext) to prompt for iOS authentication. And the supported [policies](https://developer.apple.com/documentation/localauthentication/lapolicy) only allow you to determine whether-or-not to allow passcode fallback.

robbiehanson:
Addresses issue #455

dpad85 (approved):
LGTM

PR #481, opened 2023-11-30
(ios) Fixing crash in MinerFeeSheet

robbiehanson:
This crash occurs when `estimateFeeForSpliceOut` is called while Offline. In which case the function actually returns nil (instead of a Pair).

### DIFF
--- a/phoenix-ios/phoenix-ios/views/send/MinerFeeSheet.swift
+++ b/phoenix-ios/phoenix-ios/views/send/MinerFeeSheet.swift
@@ -465,16 +465,23 @@ struct MinerFeeSheet: View {
 					targetFeerate: feePerKw
 				)
 				
-				let updatedFeePerKw: Lightning_kmpFeeratePerKw = pair!.first!
-				let fee: Bitcoin_kmpSatoshi = pair!.second!
-				
-				if self.satsPerByte == originalSatsPerByte {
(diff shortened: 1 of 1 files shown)


robbiehanson:
Good question. The bug affected users who happen to get disconnected at a specific time. So if the user goes to make an on-chain payment, but they're disconnected, then they'll see this:



Since the "prepare transaction" button is disabled, they can't proceed. But once they connect, they can tap the button, which will present the MinerFeeSheet. 



Previously, if the user got disconnected when this sheet was open, then changing the rate (sat/vByte) would cause a crash. It no longer crashes now. It just displays an unknown cost ("? sat" / "?.?? FIAT") since the method returns nil.

If the user closes this sheet (with an unknown cost), they go back to the first screenshot above.

There's a similar situation if the user first selects the rate (sat/vByte), and then they get disconnected before tapping "Pay":



(Because the user can tap the "edit" button next to the miner fee to redisplay the MinerFeeSheet.)

But the end result, that we really cannot prevent, is that the user can have the MinerFeeSheet open, and can be disconnected.

For now we don't do much in this situation. The user sees "? sat", and that's about it. When they return to the main screen, however, they can see "waiting for internet". And that's actually a button, which displays the popover window:



What do you think ? Should we display something more in the MinerFeeSheet ?

Issue #468, opened 2023-11-14
Display LNURL Error Reason instead of "service offline" message


dr-erych:
1️⃣ Scan or paste an LNURL that returns an error response according to the [LNURL Spec](https://github.com/lnurl/luds/blob/luds/03.md#wallet-to-service-interaction-flow) (also applies to payRequest etc.) 

Example: [LNURL1DP68GURN8GHJ7MR9VAJKUEPWD3HXY6T5WVHXXMMD9AMKJARGV3EXZAE0V9CXJTMKXYHKCMN4WFKZ76TJXDVYVC6RFQM8W52Y8PU9QE2PXDZKXKR34H0W77](https://legend.lnbits.com/withdraw/api/v1/lnurl/ir3XFcCH6wQD8xPeA3EcXq)  

<img src="https://github.com/ACINQ/phoenix/assets/13932145/b8220d86-fe89-4cd8-82e8-010341d0bb18" width=170>

Returns:  
`{"status":"ERROR","reason":"Withdraw is spent."}`

2️⃣ The Phoenix wallet (at least on iOS) displays "The service (...) appears to be offline, or they have a down server", which is not the case.  
It would be helpful to the user to actually know the "reason" from the LNURL response.

<img width="175" alt="image" src="https://github.com/ACINQ/phoenix/assets/13932145/1b2e3292-22d7-4650-89f5-595a4582b596">

dpad85:
Thanks for reporting this issue.

It seems specific to the iOS app. @robbiehanson I think we'll fallback to an "service offline" message if the error is not unreadable/cannot connect : https://github.com/ACINQ/phoenix/blob/3d122f583d00fea107d8e56bc26fd2f1a139216c/phoenix-ios/phoenix-ios/views/send/SendView.swift#L241-L248 

On Android we should be ok:
https://github.com/ACINQ/phoenix/blob/3d122f583d00fea107d8e56bc26fd2f1a139216c/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/ScanDataView.kt#L298-L303

Though on Android we don't have a special Lightning address case (user unknown by this service), which is a nice feature.

PR #482, opened 2023-12-01
Display LNURL Error Reason

robbiehanson:
Fixes issue #468 

Initially this looked like a bug specific to iOS. But I found another problem in the shared kotlin layer:

The error returned from the referenced [lnurl](https://legend.lnbits.com/withdraw/api/v1/lnurl/ir3XFcCH6wQD8xPeA3EcXq):
* is `LnurlError.RemoteFailure.Code(code = 404)`
* but it should be `LnurlError.RemoteFailure.Detailed(reason = "Withdraw is spent.")`

The problem is that we weren't properly following the [LUD-01](https://github.com/lnurl/luds/blob/luds/01.md) spec:

> HTTP Status Codes and Content-Type:
> 
> Neither status codes or any HTTP Header has any meaning. Servers may use whatever they want. Clients should ignore them (and be careful when using libraries that treat responses differently based on headers and status codes) and just parse the response body as JSON, then interpret it accordingly.

So even though the server returns a 404, since it also returns a valid JSON (that matches the spec with `status=error` & `reason=string`), we should be returning the given reason as a `RemoteFailure.Detailed`.

### DIFF
--- a/phoenix-ios/phoenix-ios/Localizable.xcstrings
+++ b/phoenix-ios/phoenix-ios/Localizable.xcstrings
@@ -18280,8 +18280,12 @@
         }
       }
     },
+    "The invoice is not for (chainMismatch.expected.name)" : {
+      "comment" : "Error message - scanning lightning invoice"
+    },
     "The invoice is not for (reason.expected.name)" : {
       "comment" : "Error message - scanning lightning invoice",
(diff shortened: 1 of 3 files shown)


dpad85 (approved):
Good catch with the http error code, it was in fact the main issue.

Issue #342, opened 2023-02-17
Improved support for Advanced data protection mode


robbiehanson:
Starting with iOS 16.2, iPadOS 16.2 and macOS 13.1, users can enable Advanced Data Protection to protect **select** iCloud data with end-to-end encryption.

However, according to the [docs](https://support.apple.com/en-gb/HT202303):

> Third-party app data stored in iCloud is always encrypted in transit and on server. When you turn on Advanced Data Protection, third-party app data stored in iCloud Backup and CloudKit encrypted fields and assets are end-to-end encrypted.

This would suggest that any data stored in CloudKit is NOT encrypted end-to-end unless the developer explicitly tags the database column as an encrypted column.

Initial Tasks:

- [x] Verify understanding by enabling Advanced Data Protection, and then checking your account data within CloudKit Dashboard
- [ ] Explore migration solutions for moving to encrypted columns

robbiehanson:
**Before** enabling Advanced Data Protection:

I can see my data within the [CloudKit Dashboard](https://icloud.developer.apple.com/dashboard/). That is, from the dashboard I can select "Act as iCloud account" and then login using any of my developer accounts to view the associated CloudKit data (for the Phoenix app).

**After** enabling Advanced Data Protection:

I can NOT see my data within the dashboard. When I select "Act as iCloud account" and login, I'm told I need to enable "Access iCloud Data on the web". After enabling this option, I then get a cryptic error message when I try to login. I don't know if this is an Apple Bug, or if it's "working as intended".

However, there is more information available online now. And multiple developers have confirmed that, when Advanced Data Protection is enabled, data within CloudKit is NOT encrypted E2E, except for encrypted fields & assets.

Further, there is no API to determine whether or not the user has Advanced Data Protection enabled.

robbiehanson:
`CKRecord.encryptedValues` [requires](https://developer.apple.com/documentation/cloudkit/ckrecord/3746821-encryptedvalues) iOS 15+

We are currently supporting iOS 14. But there's no reason to continue supporting iOS 14 since every device that can run iOS 14 can also run iOS 15. That is, when Apple released iOS 15, they didn't drop support for even a single iPhone or iPad device.

PR #485, opened 2023-12-07
(ios) Apple's "advanced data protection" + seed backups

robbiehanson:
Apple recently added end-to-end encrypted data for iCloud via [Advanced Data Protection](https://support.apple.com/guide/security/advanced-data-protection-for-icloud-sec973254c5f/web):

<br/><br/>
When I first heard about this, I was under the impression that **ALL** iCloud data would be encrypted E2E (except mail, calendar & contacts). However, this is false. When enabled, all of Apple's **own** apps get E2E protection, but 3rd party apps ... **nope**! They have to opt-in themselves:

> Advanced Data Protection also automatically protects CloudKit fields that third-party developers choose to mark as encrypted, and all CloudKit assets.

> CloudKit Record fields must be explicitly declared as “encrypted” in the container’s schema to be protected, and reading and writing encrypted fields requires the use of dedicated [APIs](https://developer.apple.com/documentation/cloudkit/ckrecord/3746821-encryptedvalues).

So with this PR, we are opting in.

### Ramifications

Advanced data protection is not enabled by default. It's opt-in, and it's a relatively hidden setting. To enable it requires the user to setup either a "recovery key" or "recovery contact".

The recovery key option is similar to backing up a bitcoin seed. The user is given a 28-character hexadecimal string, and are told they're responsible for saving it.

The recovery contact option allows them to choose 1 or more trusted contacts:

> A recovery contact can generate a code from their Apple device to help you get your data back.

The user can pick multiple options (e.g. recovery key plus multiple recovery contacts).

But throughout the setup process, the user is regularly reminded about data loss.

advancedDataProtection-iOS

So if the user loses access to their Apple account, and they lose all their recovery options, then they've basically lost all their iCloud data (photos, files, notes, etc).

Apple can help the user regain access to their account, but only the non-encrypted data would be available. Such as email, calendar, contacts ... and unencrypted CloudKit records. 

In my opinion, if a user explicitly enables "advanced data protection", they would probably assume that an iCloud backup of their recovery phrase (via Phoenix, or any other wallet) would be E2E encrypted. They might be wrong. But they would be excused for believing so, because most people do.

_In fact, if you watch any tutorial on setting up Advanced Data Protection, most people say something like "all your iCloud data is E2E encrypted, except mail, calendar & contacts." ([example](https://www.youtube.com/watch?v=HxoAwnRFwJ0)) So the confusion is common._

For this reason (and others) I think the responsible choice is to opt-in to E2E encryption.

### Technical notes

The record we're now backing up in the cloud looks like this:

```json
{
  "name": "Optional backup name user can enter",
  "language": "en",
  "encrypted": {
    "mnemonics": "indoor fox shiver ..."
  }
}
```

In other words, in it's current state, this PR encrypts just the recovery phrase, and not the other metadata. Requesting feedback about this.

### DIFF
--- a/phoenix-ios/phoenix-ios/sync/SyncSeedManager.swift
+++ b/phoenix-ios/phoenix-ios/sync/SyncSeedManager.swift
@@ -11,16 +11,17 @@ fileprivate var log = LoggerFactory.shared.logger(filename, .trace)
 fileprivate var log = LoggerFactory.shared.logger(filename, .warning)
 #endif
 
-fileprivate let record_column_mnemonics = "mnemonics"
-fileprivate let record_column_language = "language"
 fileprivate let record_column_name = "name"
+fileprivate let record_column_language = "language"
+fileprivate let record_column_mnemonics_deprecated = "mnemonics" // record[key]
+fileprivate let record_column_mnemonics_encrypted = "mnemonics2" // record.encryptedValues[key]
 
 struct SeedBackup {
 	let recordID: CKRecord.ID
-	let mnemonics: String
-	let language: String
-	let name: String?
 	let created: Date
+	let name: String?
+	let language: String
+	let mnemonics: String
 }
 
 enum FetchSeedsError: Error {
@@ -140,16 +141,20 @@ class SyncSeedManager: SyncManagerProtcol {
 							
 							if case .success(let record) = result {
 								
-								if let mnemonics = record[record_column_mnemonics] as? String,
+								let mnemonics =
+									record.encryptedValues[record_column_mnemonics_encrypted] as? String ??
+									record[record_column_mnemonics_deprecated] as? String
+								
+								if let name = record[record_column_name] as? String?,
 									let language = record[record_column_language] as? String,
-									let name = record[record_column_name] as? String?
+									let mnemonics
 								{
 									let item = SeedBackup(
 										recordID: recordID,
-										mnemonics: mnemonics,
-										language: language,
+										created: record.creationDate ?? Date.distantPast,
 										name: name,
-										created: record.creationDate ?? Date.distantPast
+										language: language,
+										mnemonics: mnemonics
 									)
 									
 									publisher.send(item)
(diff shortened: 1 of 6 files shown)


robbiehanson:
Fixes issue #342

phoenix-ios/phoenix-ios/sync/SyncSeedManager.swift
```diff
@@ -350,9 +576,9 @@ class SyncSeedManager: SyncManagerProtcol {
 						recordID: recordID()
 					)
 					
-					record[record_column_mnemonics] = recoveryPhrase.mnemonics
-					record[record_column_language] = recoveryPhrase.languageCode
 					record[record_column_name] = uploadedName
+					record[record_column_language] = recoveryPhrase.languageCode
+					record.encryptedValues[record_column_mnemonics_encrypted] = recoveryPhrase.mnemonics
```
dpad85, line 581: What happens if the user has not enabled "Advanced data protection"? Does `CKRecord.encryptedValues` falls back to a non-encrypted value?

dpad85 (commented):
> if a user explicitly enables "advanced data protection", they would probably assume that an iCloud backup of their recovery phrase would be E2E encrypted.

That makes sense, but if we know from within Phoenix that this advanced option was enabled in iOS, then we should remind them somewhere in the UI that this option also applies to their seed backup. 

It's likely that users enable this seed iCloud backup option for peace of mind. It's a trade-off that ensures they ever won't lose their funds because they forgot "yet another" password, and trusting Apple to not access their seed. By enabling Advanced data protection on this, this expectation is not valid anymore, so they should know it.

> this PR encrypts just the recovery phrase, and not the other metadata

Sounds good to me, the other fields in the record are not important.

robbiehanson:
> but if we know from within Phoenix that this advanced option was enabled in iOS

There's a catch. Apple does not tell us whether or not "advanced data protection" is enabled. So Phoenix has no way of knowing. :slightly_frowning_face:

And also, if the user enables this option, Apple makes it really easy to disable (which simply uploads the private keys to Apple's servers). So if it's enabled today, it could theoretically be disabled tomorrow.

dpad85:
> Apple does not tell us whether or not "advanced data protection" is enabled

Ok then we just need to add a short disclaimer about "Advanced data protection" in Settings > Recovery Phrase > iCloud backup, below the `Your recovery passphrase will be stored...` message, and I think we should be good.

robbiehanson:
Here's what I came up with:



So the user has to agree to 1 of the legal sections. And we use ~~strikethrough~~ to explain the implications of "advanced data protection".

The (i) button points the user to Apple's [official page](https://support.apple.com/en-us/108756)

dpad85 (commented):
I'm not 100% sure about the strikethrough though, it's a bit unusual, like there's a problem.

Another option could be: leave the UI as it was, but add a message at the bottom that says 1) If enabled, Advanced data protection applies to the seed backup and you _should_ be safe from Apple/Government but 2) make sure your keep your iCloud recovery method up-to-date.

robbiehanson:
Updated design:

dpad85 (approved):
🚀

PR #490, opened 2023-12-15
Updated Czech Translation

MasterixCZ:
I have corrected and updated the Czech translation of the new feature.


dpad85 (approved):
Thanks for the update @MasterixCZ !

PR #494, opened 2023-12-20
(iOS) Strings localized to es-419 (Spanish for Latin America and the Caribbean)

BitByBitByBitByBit:
This is the localization to es-419 (Spanish for Latin America and the Caribbean).

The English strings have been translated from scratch as the current Spanish translation could be more suitable for Spain.


dpad85:
Thanks for this large update! We'll review it asap.

BitByBitByBitByBit:
I have updated `strings.xml` and `important_strings.xml` for the Android localization too (I modified the PR title accordingly, let me know if this needs to be done differently).

dpad85 (approved):
LGTM, thanks!

PR #509, opened 2024-01-22
Upgrade to Kotlin 1.9

dpad85:
This PR upgrades kotlin to version 1.9, which has side effects :
- upgrading gradle ;
- upgrading several libraries, including `secp256k1`, `lightning-kmp` ;
- updating the gradle config ;
- removing obsolete libraries and cleaning up code (see `LogMemory`).

It also brings several improvements:
- possible better performance, see https://github.com/ACINQ/lightning-kmp/pull/565 for details ;
- let us update SQLDelight (in a future PR) ;
- let us switch to Kermit for logging (in a future PR) @robbiehanson 

### Libraries update

- ktor upgraded to 2.3.7
- kotlinx.coroutines to 1.7.3
- kotlinx.serialization to 1.6.2

⚠️ `lightning-kmp` is using a development version.

### About `iosSimulatorArm64`

The `iosSimulatorArm64` target has been removed for now as it's not yet supported by `tor-mobile-kmp`.

### Removing `LogMemory`

The `kodein-memory` library has also been removed. This library was used to write the logs to disk in the shared module. However we were using an old version, and we are already planning to move to another log library and to delegate writing logs to disk to the application layer (*).

So it's not worth spending time upgrading this library ; instead this dependency is removed, and the code has be cleaned up:

- removed `LogMemory`
- removed usage of APIs from this lib (freeze calls)
- removed most code from the `LogConfiguration` controller, which will have to be removed altogether at some point

(*) I think that delegating writing logs to disk should help with performances.


phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/db/payments/MetadataTypes.kt
```diff
@@ -289,14 +288,6 @@ data class WalletPaymentMetadataRow(
             && user_notes == null
     }
 
-    /**
-     * This function exists because the `freeze()`
-     * function isn't exposed to iOS.
-     */
-    fun copyAndFreeze(): WalletPaymentMetadataRow {
-        return this.freeze()
-    }
```
dpad85, line 298: @robbiehanson Can you check that this method was not used anymore - since we are using the new memory model, there's no need to freeze objects anymore.

phoenix-shared/src/iosMain/kotlin/fr/acinq/phoenix/db/CloudKitDb.kt
```diff
@@ -679,7 +678,7 @@ class CloudKitDb(
         //
         return metadata.mapValues { (_, metadataRow) ->
             metadataRow.copy(
-                recordBlob = metadataRow.recordBlob.copyOf().freeze()
+                recordBlob = metadataRow.recordBlob
```
dpad85, line 681: Same as above, @robbiehanson freezing objects should not be needed anymore.

pm47 (approved):

PR #512, opened 2024-01-24
Use AppleSymbol instead of SVG

robbiehanson:
I converted the SVG icon into an AppleSymbol, which now matches the other standard iOS icons in size and thickness.

We only do this on iOS 17+, since it doesn't work properly on earlier iOS versions.


dpad85 (approved):
Looks good! Tested on 16.4 and 17.2.

If I understand correctly, you'll have to maintain 2 files for each icon?

robbiehanson:
> If I understand correctly, you'll have to maintain 2 files for each icon?

It's really just one image. When you drag that image into Xcode, and add it to an Assets Library, then Xcode automatically creates the JSON file for it's own internal bookkeeping.

Issue #507, opened 2024-01-18
UX: Subtle Max Fee Warning -> Prominent Max Fee Suggestion


kilrau:
We are seeing not few failed swaps at Boltz that turned out to be due to user's too low `max fee amount` limit in their Phoenix application that blocked the splice-in. The default of 5000 sats is just not nearly enough these days.

![image](https://github.com/ACINQ/phoenix/assets/19181985/13a34768-4533-4980-a15a-7c24d2652594)

Since we are ususally sending these kind of screenshots to make users aware that there is such setting in the first place (literally everyone we talked to overlooked the warning), how about moving from this rather subtle warning to a more prominent even *screen blocking* suggestion to "raise max fee amount to X according to current estimates, otherwise this payment is likely to fail" to make splice-ins more likely to succeed? The fee estimate is even shown in the max fee amount menu, it's just most don't even get to that screen. It should be a one button tap to take over the fee limit suggestion.

![image](https://github.com/ACINQ/phoenix/assets/19181985/c2ff8b7d-95cf-44e9-a0aa-949a65df8528)

The user should be able to deny this max fee limit suggestion but then it's crystal clear at least that the payment for this invoice amount is likely to fail and folks don't end up in our support channels blaming us :v:

dpad85:
This issue is fixed on Android with release 2.1.3, but not on iOS yet. I will update this ticket once iOS is ready.

PR #514, opened 2024-02-01
Make fee warning more prominent

dpad85:
See https://github.com/ACINQ/phoenix/issues/507 for context

The message warning users that the payment they intend to receive may fail was not visible enough. It was also quite long, while not providing enough information nor obvious solutions. This PR attempts to fix that.

### Warning is now a one-liner just below the QR code

image

### Clicking on the warning opens a dialog with more information/solutions

Note: for obvious cases, this dialog is opened immediately, even if the user has not clicked on it.

image

image

### also show a "Fee expected" message

This lets the user know that they should expect a fee, even if it is within the configured bounds. They may prefer add liquidity instead.

image

@robbiehanson I added a helper method in `LocalChannelInfo` to get the real receive balance from the channels state flow, to only take into account the relevant channels - e.g., ignore channels which are booting. This method returns null when the liquidity is unknown at the moment, so you know when warning should be displayed, or not.

See https://github.com/ACINQ/phoenix/blob/e1ed3fd538402a164ed19d7983513fe061ef4ef0/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/receive/ReceiveLightningView.kt#L332-L455 for the warning logic I have been using.


dpad85:
I also considered having a "Fix me" button (in the channels management screen) that would just change the configuration to whatever works, but it would take time to design properly and avoid footgunning users. So if we do it, it should be done in a later PR.

pm47 (approved):

dluvian (commented):
I never heard "Transaktion auf der Kette". I'd go with "on-chain" which is also used in other german translation strings in Phoenix

PR #516, opened 2024-02-09
Support taproot addresses and rotation

dpad85:
🚧  This PR is a work-in-progress. It depends on updates on the secp256k1 library [(see link for details)](https://github.com/ACINQ/lightning-kmp/pull/563) which is not expected to arrive soon. So this PR will stay in draft for some time. Most of the wording is a WIP.

### Support for taproot addresses

Phoenix now shows a taproot address by default for swap-ins (on-chain deposits to Phoenix). Taproot addresses are more private (observers won't know funds sent there are for LN channels) and are cheaper (~15% less fee, and once channels also support taproot, it could be up to 30% less expensive in total).

image

### Backward compat with legacy addresses

Phoenix will still be able to swap funds sent on the legacy address, even if it's not the preferred way. User can also switch back to the legacy address for deposits from wallets/services that don't support taproot.

image
image
image

### Address rotation

Taproot addresses are rotated, which also improves privacy besides the taproot features - see https://github.com/ACINQ/lightning-kmp/pull/584 for details. 

It means that Phoenix will synchronise its view of the swap-in wallet at startup to find the last unused address, and avoid address reuse. This synchronisation can take time (dozens of seconds if many addresses), during which the UI should still display an address (for good UX).

@robbiehanson To do so, the last known unused index should be stored in the preferences, and used at startup to provide an address to the UI:

```kotlin
keyManager.swapInOnChainWallet.getSwapInProtocol(index).address(chain)
```

Then listen to `Peer.swapInWallet.swapInAddressFlow` and use that address at collect (which should be the same address as calculated before, unless a payment was made while the app was off).

### Show used addresses

The user can check his addresses in Settings > Wallet info > Swap addresses.

image

image

These addresses are exposed in `Peer.swapInWallet.wallet.walletStateFlow`. Note that the legacy address is contained in that flow (using a `AddressMeta.Single` meta).

Also, this flow only exposes the adresses that have been used, + the first unused address. It does not list other unused addresses.


dpad85:
Note that this PR will have to be rebased on kotlin 1.9.

sstone:
side-note: The phoenix-legacy app cannot compute swap-in addresses for the new taproot-based protocol, but it should be easy to implement if needed:

- update`bitcoin-lib` 0.18.3 and add `bitcoin-kmp` 0.17.0 and `secp256k1-kmp` 0.14.0
- update `eclair-core`  v0.4.23-android-phoenix and add a method to compute new swap-in addresses 

This will make the old version of `eclair-core` used by the legacy app depend on kotlin 1.9 but I understand that it is not a problem as the legacy app is already using kotlin 1.9.

robbiehanson (approved):

PR #522, opened 2024-02-27
Monitor and settle in-flight payments when the app is in the background 

dpad85:
When a payment is pending, then times out, the app needs to be online and connected to the peer to negotiate the payment failure. If the app goes off while a payment is pending, and then does not connect once it times out, the peer is forced after a while to unilaterally close the channel.

This issue is already taken care thanks to the silent push notifications sent by the peer that will "wake up" Phoenix remotely, so that it can settle these pending payments. However, this does not work all the time and there still may be force closures.

This PR schedules a task that runs when the app is in the background/off and there are in-flight payments.

This task consists in simply starting the node, which connects to the peer and wait for any pending payments ready to be finalised to be finalised. Once all payments are settled, or after 30s (iOS) or 2 minutes (Android) the task aborts.

This tasks runs on both the iOS and the Android app (although on iOS, the task is merged with the watchtower, since the OS allows us to schedule only 1 `BGAppRefreshTask`). 

On Android, we also display a badge in the home screen that counts in-flight payments to make the user aware that a payment is in-flight.

Screenshot 2024-02-27 at 19 11 06


robbiehanson (approved):

phoenix-android/src/main/res/values-de/important_strings.xml
```diff
@@ -30,6 +30,9 @@
     <string name="notif_pending_settlement_title">Bitte öffnen Sie Phoenix</string>
     <string name="notif_pending_settlement_message">Eine eingehende Zahlung steht aus.</string>
 
+    <string name="notif_inflight_payment_title">Eine Zahlung steht an</string>
+    <string name="notif_inflight_payment_message">Starten Sie Phoenix, damit die Zahlung irgendwann abgeschlossen werden kann.</string>
```
dluvian, line 34:
```suggestion
    <string name="notif_inflight_payment_message">Starten Sie Phoenix, damit die Zahlung abgeschlossen werden kann.</string>
```
Or replace the `"irgendwann"` with `"rechtzeitig"`

dpad85: Thanks @dluvian

PR #530, opened 2024-03-06
(ui-android) Enable redacted balance mode in payments

### LINKED ISSUE #438: bug: total money not hidden when entering the amount to send
I can hide the sats in my account, if I tap on my money. But my money is not hidden, if I scan an invoice from someone and have to enter, how much I want to pay. On the top right corner, my total money is shown.

A solution to this would be to also just show stars there and be able to unhide it, if I tap on it.

ndungudedan:
Resolves #438 

When a user is sending a payment, their balance is displayed on the top right in fiat or sats. This commit adds redacted mode which shows **** if the user desires. Clicking on the balance toggles the display modes. This functionality is similar and in **sync** with the home account balance view.

20240306123714-ezgif com-crop


dpad85 (approved):
LGTM, thanks!

Issue #533, opened 2024-03-15
[UX IMPROVEMENT] Discriminate between error messages when trying to pay invoices


PaulADW:
The app does not discriminate between "fixed amount invoices" and "variable amount invoices". 
This leads to much confusion for users. 

Here's an exemple : I send an invoice for 159 069 sats, the person that has to pay it copies and pastes the invoice in the app, here's what they see : 

![WhatsApp Image 2024-03-15 at 12 43 17](https://github.com/ACINQ/phoenix/assets/50180437/84bbcc24-ccdc-4474-9113-0c68d7b4b916)

They get the possibility to enter an amount, even though the invoice they received is for a fixed amount. That is a bad UX. This is because the error message does not discriminate between Fixed amount invoices and Variable Amount Invoices. 

The UX should say something like "This invoice is for 159 069 sats (98,61€), you don't have enough funds to pay it". Also you shouldn't be able to edit the amount you want if you're trying to pay a Fixed Amount Invoice, it doesn't make sense. 

You guys should write a rule that would be like "IF INVOICED RECEIVED IS FIXED AMOUNT AND USER BALANCE IS INFERIROR TO INVOICE, THEN DISPLAY ERROR MESSAGE "YOU DONT HAVE ENOUGH MONEY TO PAY THIS INVOICE"

Here's how the user reacts with the current UX : 

<img width="330" alt="image" src="https://github.com/ACINQ/phoenix/assets/50180437/1ea831bd-3353-40ed-ab4c-55f6f9942d15">

Shouldn't be hard to fix and make it much more user friendly.

PR #538, opened 2024-03-24
[Android] - Disable editing amount when paying an invoice with a defined amount

ndungudedan:
This PR resolves issue #533.

I added a simple check where if the invoice amount is null, the input will be enabled and editable when paying the invoice.

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/SendLightningView.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/SendLightningView.kt
@@ -61,6 +61,7 @@ fun SendBolt11PaymentView(
         topContent = {
             AmountHeroInput(
                 initialAmount = amount,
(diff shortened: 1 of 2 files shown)


dpad85 (approved):
Being able to overpay an invoice with a requested amount is a feature:
- lets you tip someone (e.g., a restaurant)
- adds some privacy (see BOLT04 and overpayments)

So we should not just disable that.

However, there are services that don't handle overpayments well, causing serious issues for users. Being able to manually edit the amount should be an option (opt-in) to avoid footgun. And there should be a tip button (already exists in iOS, todo on Android) in the normal case.

Let's merge this, and we'll iterate to add these features ^.

ndungudedan:
> Being able to manually edit the amount should be an option (opt-in) to avoid footgun. And there should be a tip button (already exists in iOS, todo on Android) in the normal case.

I'm gonna take this up, and iterate on the features.

PR #539, opened 2024-03-28
Add fee/liquidity disclaimers in several screens

dpad85:
This PR adds new confirmation or disclaimer messages in several screens. The purpose is to clarify how liquidity works, especially with relation to splicing (in or out) to avoid mismanagement of the wallet.

The wording of these messages is a WIP.

# Requesting liquidity

The message in the lease duration tooltip has been updated. Clicking on Review displays a confirmation message. The user needs to acknowledge this message to proceed further and get the liquidity.

image

image

# Splice-out

When finalising a splice-out (sending on-chain), a message is displayed to remind the user that sending funds on-chain mechanically diminishes the capacity of the channel. This might not be obvious, and some users may instead use a Lightning payment instead.

The capacity disclaimer is shown every time a splice-out is done, except if the user has checked the "don't show again" box.

Example with 2 warning before payment: low feerate warning + capacity disclaimer:
- trying to use a low feerate
- click on pay
- 1st message about low feerate
- 2nd message about effect on capacity
- confirming => executing the splice

image

image

image


robbiehanson:
&nbsp;&nbsp;&nbsp;&nbsp;

robbiehanson (approved):

PR #551, opened 2024-04-11
(ios) Revert: Integrate Touchlab's SKIE

robbiehanson:
We were experiencing app freezes in production. The root cause of the problem was Touchlab's SKIE. In particular, an unexpected quirk of the SKIE implementation.

Consider this Swift code:
```swift
Task { @MainActor in
   self.fetched = try await paymentsManager.fetcher.getPayment(
      row: row,
      options: PaymentCell.fetchOptions
   )
}
```

Here we're calling a `suspend` function in Kotlin (`getPayment`) from Swift. Now it has always been the case this must be done on the main thread (e.g. `MainActor`). In fact if you tried from a background thread, Kotlin would throw an exception. It has also been the case that `getPayment` executes on the main thread.

In recent versions of Kotlin, this main-thread restriction could be [removed](https://kotlinlang.org/docs/native-ios-integration.html#calling-kotlin-suspending-functions) with a flag.

It appears that SKIE automatically sets this flag. From their [docs](https://kotlinlang.org/docs/native-ios-integration.html#calling-kotlin-suspending-functions)

> Another limitation is that these suspend functions can only be called from the main thread. This limitation can be solved without SKIE by passing a special configuration flag to the compiler. However, SKIE adds this support out of the box.

The end result is that multiple instances of `getPayment` are running simultaneously on different background threads, our built-in Cache is getting corrupted, and the linked-list ends up with cycles leading to infinite loops in certain situations.

It's relatively easy to solve this one particular bug. But the bigger issue is that I don't know what other issues this problem might cause. And time is needed to further research the issue.


dpad85 (approved):
> The end result is that multiple instances of getPayment are running simultaneously on different background threads, our built-in Cache is getting corrupted, and the linked-list ends up with cycles leading to infinite loops in certain situations.

> It's relatively easy to solve this one particular bug. But the bigger issue is that I don't know what other issues this problem might cause. And time is needed to further research the issue.

Makes sense, the SyncTxManager could be another cause, and there may be others. Will make a build ASAP.

PR #560, opened 2024-05-10
(ios) Custom PIN as an option for “app access”

robbiehanson:
We've had many requests for a "custom PIN" option. That is, the ability to set a PIN/passcode that's specific to Phoenix, and possibly different than the system PIN/passcode.

This PR allows us to explore how that feature would operate in practice.

On the surface it's quite simple:

A new PIN option is added to the "app access" section. The user can choose any of the following options:
* Face ID only
* Face ID + system PIN fallback
* Face ID + custom PIN fallback
* Custom PIN only

The custom PIN length is 6 digits. (This matches the iOS default PIN length.)

And the PIN will be presented to the user from within the Lock screen:

&nbsp;&nbsp;

The usual suspects are there:

* if the user enters the wrong PIN multiple times, there's a "slowdown" timer that kicks in (anti-brute-force)
* Invalid PIN attempts are stored in the keychain, so even if they relaunch the app, the slowdown timer is still active
* when the user creates their PIN, they have to enter it twice to confirm

However, there are some iOS specific "things to consider":

The "system PIN fallback" was added in PR #478. Recall that when we prompt for "system authentication" we basically only have 3 options on iOS:

* Face ID (biometrics) only
* Face ID (biometrics) + system PIN fallback
* Face ID (biometrics) + custom (app specific) fallback

In other words, prompting for the system PIN is not an option. It's only available as a fallback option (after biometrics fails).

So the "system PIN fallback" option was primarily motivated by user expectations. But was also **partly** motivated by user reports of broken Face ID (due to broken hardware). It was thought that, if Face ID breaks, then users would still be able to access Phoenix via the system PIN. However, recent reports have cast doubt on that assumption.

> When I go to Settings -> Face ID & code, I get the message "A problem was detected with the TrueDepth camera. Face ID has been deactivated"

In other words, in certain situations when Face ID breaks, the system automatically disables it on a system level. Making it no longer accessible from within apps. Meaning even with the "system PIN fallback" users would be locked out of Phoenix. To re-access their wallet, they would need to reinstall & restore their wallet. The custom PIN option does present an alternative here.


robbiehanson:
Is issue #441 we said:

> In all cases, the seed should be backed up before enabling any of these [app access] options. Or at least show a visible warning.

We could add that here, or make it a separate PR.

dpad85:
> We could add that here, or make it a separate PR
I think it's fine in this PR (it should not be a large change?)

phoenix-ios/phoenix-ios/security/InvalidPin.swift
```diff
@@ -0,0 +1,38 @@
...
+		let newCount = (count < Int.max) ? count + 1 : Int.max
+		let newTimestamp = Date.now
+		
+		return InvalidPin(count: newCount, timestamp: newTimestamp)
+	}
+	
+	var waitTime: TimeInterval { // TimeInterval => seconds (Double)
+		switch count {
+			case 0  : return  0.seconds()
+			case 1  : return  0.seconds()
```
dpad85, line 21: I think we could allow 3 quick failed attempts? or make the third attempt wait 10s.

Also, how long does it take for the interval to reset? E.g, I made 5 failed attempts. If I wait 1 day and try again, does the failed attempts count reset to 0?

phoenix-ios/phoenix-ios/views/compatibility/VibrationFeedback.swift
```diff
@@ -0,0 +1,36 @@
...
+	@available(iOS 17.0, *)
+	func convert() -> SensoryFeedback {
+		switch self {
+			case .error : return SensoryFeedback.error
+		}
+	}
+}
+
+struct VibrationFeedback<T>: ViewModifier where T: Equatable {
+	let type: VibrationFeedbackType
```
dpad85, line 15: When do these vibration happen? on my device (iphoneX, iOS 16.7.7) with haptic enabled in Settings > Sound & haptic, I don't get feedback when pressing PIN buttons, or when entering an invalid PIN.

phoenix-ios/phoenix-ios/views/configuration/privacy and security/app access/NumberPadView.swift
```diff
@@ -0,0 +1,230 @@
...
+				buttonPressed(identifier)
+			} label: {
+				VStack(alignment: HorizontalAlignment.center, spacing: 2) {
+					
+					Text(verbatim: identifier.rawValue)
+						.font(.system(size: 34, weight: .medium))
+					
+					Group {
+						switch identifier {
+							case ._2 : Text(verbatim: "ABC")
```
dpad85, line 132: Why are there alphabetic characters in these buttons? IIUC the PIN is digits-only, and the UI does not let me long-press PIN buttons to use alphabetic chars instead of digits. Or am I missing something?

phoenix-ios/phoenix-ios/security/InvalidPin.swift
robbiehanson: > If I wait 1 day and try again, does the failed attempts count reset to 0?

Currently, no. But that's a really good idea. I'll add it.

phoenix-ios/phoenix-ios/views/compatibility/VibrationFeedback.swift
robbiehanson: That particular feature is iOS 17 only. Although there are other ways to accomplish it on older iOS versions. I'll take a look.

phoenix-ios/phoenix-ios/views/configuration/privacy and security/app access/NumberPadView.swift
robbiehanson: Because the iOS system pin (numbers only) also displays the alphabetic characters:



I assume Apple has always done it this way to help users create easy-to-remember PINS. E.g. "dios10" => 346710

I think Android might do it too, but I'm not sure about recent versions.

phoenix-ios/phoenix-ios/security/InvalidPin.swift
robbiehanson: Done in https://github.com/ACINQ/phoenix/pull/560/commits/0218ed64c873c873dfb066b9a0ab30cae35ba790

robbiehanson:
All options are now disabled until the user performs a backup:

phoenix-ios/phoenix-ios/views/compatibility/VibrationFeedback.swift
robbiehanson: Done in https://github.com/ACINQ/phoenix/pull/560/commits/70cb899f3dce3cfc3498ae52997d84685a4d7a77

phoenix-ios/phoenix-ios/views/configuration/privacy and security/app access/NumberPadView.swift
robbiehanson: However, this is probably only the case for languages that use a Latin alphabet... I'll update the code to only show the letters in that situation.

robbiehanson: Done in https://github.com/ACINQ/phoenix/pull/560/commits/2614dfed6fa1b44b7cc2c54ad0cc3860fc8faf58

PR #561, opened 2024-05-13
(android) Backup channels and payments db

dpad85:
This PR adds a local, on-device backup mechanism for the payments and channels database, for Android. It will allow users to restore their payments history after uninstalling or resetting the wallet.

The channels and payments database are zipped and encrypted on disk in public folder. This backup file will not be erased when the app is uninstalled, and can thus can be restored manually when the user imports their wallet again in the future. It does not require new permissions, since we are using the Mediastore API to write the file, and importing the backup file is done manually by the user, using the Android file browsing framework.

This encrypted file can also be backed up using the auto Google Drive backup system, if available and enabled, which would avoid having to manually restore the local backup file.

This PR and description is a WIP.


dpad85:
Closing this PR because the automated backup system for the Android app should be based on the manual payments import/export feature added in #685.

PR #563, opened 2024-05-13
Handle new outgoing payment failures

dpad85:
This PR supports the new outgoing payment failures types added in https://github.com/ACINQ/lightning-kmp/pull/634.

There is a small impact on the payments database, as the status of failed parts are stored using a `code: Int` json (with an optional String message).

@robbiehanson we should remove the details for failed parts on iOS. Instead, show a single error message using the `OutgoingPaymentFailure.explain()` method that will pick the most pertinent error for a given payment (will return either a `FinalFailure` error, or a `Part.Status.Failure`)

### DIFF
--- a/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/db/payments/OutgoingPartStatusType.kt
+++ b/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/db/payments/OutgoingPartStatusType.kt
@@ -100,6 +100,8 @@ sealed class OutgoingPartStatusData {
 fun LightningOutgoingPayment.Part.Status.Succeeded.mapToDb() = OutgoingPartStatusTypeVersion.SUCCEEDED_V0 to
         Json.encodeToString(OutgoingPartStatusData.Succeeded.V0(preimage)).toByteArray(Charsets.UTF_8)
 
+fun LightningOutgoingPayment.Part.Status.Failed.mapToDb() = this.failure.mapToDb()
+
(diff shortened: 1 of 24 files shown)


phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/db/payments/OutgoingPartStatusType.kt
```diff
@@ -69,5 +99,22 @@ sealed class OutgoingPartStatusData {
...
+        is LightningOutgoingPayment.Part.Status.Failure.PaymentExpiryTooBig -> 5
+        is LightningOutgoingPayment.Part.Status.Failure.TooManyPendingPayments -> 6
+        is LightningOutgoingPayment.Part.Status.Failure.ChannelIsSplicing -> 7
+        is LightningOutgoingPayment.Part.Status.Failure.ChannelIsClosing -> 8
+        is LightningOutgoingPayment.Part.Status.Failure.TemporaryRemoteFailure -> 9
+        is LightningOutgoingPayment.Part.Status.Failure.RecipientLiquidityIssue -> 10
+        is LightningOutgoingPayment.Part.Status.Failure.RecipientIsOffline -> 11
+        is LightningOutgoingPayment.Part.Status.Failure.RecipientRejectedPayment -> 12
+    }
+    return OutgoingPartStatusTypeVersion.FAILED_V0 to
```
dpad85, line 118: Should use `FAILED_V1` instead of `V0` which is now obsolete.

robbiehanson (approved):

PR #564, opened 2024-05-13
Add screen to spend expired swap-ins

dpad85:
This PR adds a new screen accessible from the swap-in wallet view that lets users spend cancelled funds that have expired and are not eligible to be swapped to Lightning anymore.

Previously, users needed to perform [a manual procedure](https://github.com/ACINQ/lightning-kmp/blob/master/RECOVERY.md#pending-swap-in-transactions) to spend cancelled funds.

@robbiehanson if you have suggestions for the wording in the new screens, do not hesitate.

This PR depends on https://github.com/ACINQ/lightning-kmp/tree/spend-expired-swapin (not merged yet).

#### Tapping on Cancelled funds opens the screen to spend the cancelled funds.
image

#### The address and the feerate are set by the user.

The amount cannot be modified. It depends on what is available in the wallet.
See: `peerManager.swapInWallet.readyForRefund.balance`. 

image

#### Low feerate check

Using a feerate below the mempool.space hour estimation triggers a "low feerate" warning, like for the splice-out screen (although unlike splice-outs, Phoenix cannot splice-CPFP that transaction later on, so the message is different and the acceleration part of it is removed).

image


dpad85:
Cherry-picked `ios-564` to include the iOS changes and merged `master`.

I also added support for https://github.com/ACINQ/lightning-kmp/pull/657. This PR makes the peer's client and watcher interfaces (i.e., may not always be `ElectrumClient/Watcher`). 

In theory it means`Peer.swapInWallet` and `Peer.finalWallet` may be null, although in our case, they are always defined since Phoenix is always using Electrum. So I've added 2 extensions for Peer (see `PeerManager`): `Peer.phoenixSwapInWallet` and `Peer.finalSwapInWallet`. @robbiehanson if you used `Peer.swapInWallet` and `Peer.finalWallet` in the iOS app, you will have to switch and use these 2 extensions instead.

robbiehanson (approved):

PR #565, opened 2024-05-16
(ios) Update localizations

robbiehanson:
Existing translations updated for iOS.

The following commands were run:
```
langtool translate --source ios --lang fr
langtool translate --source ios --lang es
langtool translate --source ios --lang cs
langtool translate --source ios --lang ar
```

### DIFF
--- a/phoenix-ios/phoenix-ios/Localizable.xcstrings
+++ b/phoenix-ios/phoenix-ios/Localizable.xcstrings
@@ -3,6 +3,18 @@
   "strings" : {
     "" : {
       "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : ""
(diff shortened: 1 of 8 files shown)


dpad85 (approved):
For future localisation PRs, it would be better to split the localisation change in several commits, 1 for each language updated (within the same PR). It will make the review easier since the localisation resources are all in the same file on iOS.

I also fixed a few strings in the French localisation and I've noticed that DeepL chose to use a more informal tone for the translation (https://en.wikipedia.org/wiki/T–V_distinction), which does not work well in this setting. I suspect it did the same for the other languages. The lang tool should be updated to use the `prefer_more` option for the DeepL `formality` setting (see https://developers.deepl.com/docs/api-reference/translate), so that we get a more more formal localisation.

PR #567, opened 2024-05-23
(android) Add fee warning message in lnurl-withdraw screen

dpad85:
Withdrawals over LNURL are subject to the same liquidity constraints as regular Lightning payments, so a warning should be displayed.

It's actually even more important to have a liquidity warning here, since many LNURL services (e.g. Telegram bots) do not handle failed payments well and declare the withdrawal has been consumed, even when the payment was not received.

This PR also improves the regular Receive screen on Android, where the fee warning message dialog was opened systematically, even when the user had just seen it.

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/AppView.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/AppView.kt
@@ -273,7 +273,8 @@ fun AppView(
                         ScanDataView(
                             input = input,
                             onBackClick = { popToHome(navController) },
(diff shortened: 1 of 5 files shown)


dpad85:
Note that this PR currently depends on #563, and will need to be rebased once that other PR is merged.

dpad85:
@robbiehanson if the iOS app does not have the fee warning in the lnurl-withdraw screen either, you can fix that in this PR as well.

dpad85:
Example with a testnet withdrawal:

image


The warning dialog is the same as in the Receive screen. The clickable "Payment will fail" message is also the same.

PR #572, opened 2024-06-10
Updated Czech translation

MasterixCZ:
- New translated strings added
- Updated translations to match their English counterparts
- Fixed typos and bad translations


dpad85 (approved):
Thanks for the update.

I noticed that you also translated some of the strings related to offers which is fine but just to let you know, this update is not ready yet, so these strings will probably change in the near future.

PR #580, opened 2024-07-06
Updates to the Czech translation.

MasterixCZ:
- added translated strings about BOLT12 offers
- corrected automatic translations 
- typo correction


dpad85:
It look OK so let's keep it as-is.

PR #587, opened 2024-07-11
Updated values-cs/strings.xml

MasterixCZ:
I don't know if these strings are ready for release or if they will change, but here is the translation.


dpad85 (approved):

dpad85:
The release was built yesterday, so this fix will be added to the next one.

PR #594, opened 2024-07-15
fix: "share" translation in german

escapedcat:
"aktie" means "share" in a stock context, which is wrong.


HerrBertling:
Now this is what I call a productive coffee break 😄

dpad85 (approved):
Thanks for the fix!

PR #602, opened 2024-07-31
(ios) UI for claiming BIP353 address

robbiehanson:
Mirrors changes in #585 

Once an address is claimed, it shows up in the Receive screen:


robbiehanson:
UI modifications:

1) When the lightning address is available, we no longer display the raw offer
2) The lightning address is copied with the ₿ prefix
3) The lightning address is displayed with the ₿ prefix



4) Tapping "copy" or "share" buttons now always displays the copy/share options sheet
5) The "default" option is subtly displayed in bold

dpad85 (approved):
🚀

Issue #601, opened 2024-07-31
invoice between Android and iOS have different capital case


evansmj:
On Android, if you copy an invoice it is copied in lowercase letters.
On iOS, if you copy an invoice it is copied in UPPERCASE letters.  

In BOLT 11 it says QR codes should use UPPERCASE, but lowercase is preferable for presentation.  

I think in terms of presentation we may want the copied text to be lowercase.  

(This was discovered because a wallet was not able to handle the uppercase invoice)

---

https://github.com/lightning/bolts/blob/master/11-payment-encoding.md
* SHOULD use upper case for QR codes (see BIP-0173)

https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki
BIP-0173

```Uppercase/lowercase

The lowercase form is used when determining a character's value for checksum purposes.

Encoders MUST always output an all lowercase Bech32 string. If an uppercase version of the encoding result is desired, (e.g.- for presentation purposes, or QR code use), then an uppercasing procedure can be performed external to the encoding process.

Decoders MUST NOT accept strings where some characters are uppercase and some are lowercase (such strings are referred to as mixed case strings).

For presentation, lowercase is usually preferable, but inside QR codes uppercase SHOULD be used, as those permit the use of [alphanumeric mode](http://www.thonky.com/qr-code-tutorial/alphanumeric-mode-encoding), which is 45% more compact than the normal [byte mode](http://www.thonky.com/qr-code-tutorial/byte-mode-encoding).
```

evansmj:
This is now merged with #610

PR #604, opened 2024-08-01
iOS: Share invoices lowercased

evansmj:
According to BIP-0173, for presentation, lowercase invoices are preferable. The QRCodes are correctly encoded as uppercase here, but the text is also pasted as uppercase. Android copies the text as lowercase and since presentation is preferred to be lowercase I set the text to lowercase here.

Issue: https://github.com/ACINQ/phoenix/issues/601


robbiehanson:
Good find and good fix.

Notes: This PR will cause conflicts with #602, which is almost ready to be merged. I can circle back and patch this PR after 602 is merged.

robbiehanson:
superseded by #610

PR #610, opened 2024-08-12
iOS: Share invoices lowercased

robbiehanson:
This is just PR #604 with conflicts resolved.


phoenix-ios/phoenix-ios/views/receive/LightningDualView.swift
```diff
@@ -711,7 +711,10 @@ struct LightningDualView: View {
 			
 			offerStr = offerString
 			if activeType == .bolt12_offer {
-				qrCode.generate(value: offerString)
+				qrCode.generate(value: QRCodeValue(
+					original: offerString,
+					rendered: offerString // is this supposed to be uppercase ?
```
evansmj, line 716: I am checking with the bolt12 spec authors whether bolt12 QR Codes should be UPPERCASE similar to bolt11 and will report back

dpad85: Thanks, we can revisit later, it may take some time to get a final answer from the specs.

dpad85 (approved):

PR #618, opened 2024-09-02
(android) Add daily-connect worker

dpad85:
This PR adds a background worker to the Android app, scheduled to run roughly every day. It simply connects to the LSP, wait for 60s then shuts down. Its purpose is to settle pending actions, for example pending payment that were missed by the inflight-payments worker, and thus avoid force-closings.

Some utility code has been factorised with the inflight-payments worker.


dpad85:
Note, this worker does not replace the channels-watcher that runs every ~3 days, as the channels-watcher does not connect to the peer (on purpose).

Issue #552, opened 2024-04-15
Remove animated navigation on iOS


dpad85:
I think in general we should remove all animated navigation, and instead directly navigate to the correct screen:
- animated navigation take control away from the user (which feels weird) and waste time
- probably require complex code, which cause issues & maintenance

For example:

When tapping on a pending deposit amount in the Home screen, the app redirects to the "Swap-in wallet" view ; that part works fine.

But then, when tapping on "Configure fee settings", the app navigates (with an animation) to the Settings screen. Tapping on the Back button navigates to the Home screen, then to the Settings screen again.

robbiehanson:
I agree. Note that this probably requires full adoption of [NavigationStack](https://developer.apple.com/documentation/swiftui/navigationstack), which is iOS 16+. Which means dropping support for iOS 15. With iOS 18 coming out soon, it's probably time...

robbiehanson:
Done via #623

PR #623, opened 2024-09-09
(ios) Upgrading to path-based navigation for iOS 17+

robbiehanson:
In iOS 16 Apple introduced NavigationStack & modern path-based navigation. However, Apple's implementation remained horribly broken (see #333) until the release of iOS 17. 

This PR:
- adds full support for path-based navigation in iOS 17+
- remains backward compatible with iOS 16
- drops support for iOS 15

Fixes issue #552 (Remove animated navigation on iOS).

For example, if you navigate to:
Settings > Contacts List > Some Contact

and then tap the "send payment" button:

- the app used to slowly animate one-screen-at-a-time back to the Home screen, and then animate to the Send view
- now, with path-based navigation, it just jumps immediately to the Send Payment screen.

Note that we're still jumping from path `/Settings/ContactsList/ViewContact` to `/SendPayment`. That is, we're not **appending** the SendPayment view outside of it's normal location.

However, this change should be strictly better than what we have now. More reliable navigation, minus the goofy animations. At least on iOS 17+.

Note: Appending a view (to the current path) might be preferable in some situations. But I'll have to implement those cases in a future PR. The way NavigationStack is implemented, you can easily cause a crash if you end up with the same view on the stack twice. So either I need to prevent such loops, or I need to do some re-org.


dpad85 (approved):
Tested on an iPhoneX with iOS 16.7. Dropping support for iOS 15 is not great, but there are only a few users who are still on iOS 15.

I'm a bit surprised at the state of navigation in SwiftUI, their solution seems very verbose and complex compared to Compose in Android.

PR #628, opened 2024-09-18
Upgrade to Kotlin 2.0.10

dpad85:
This PR upgrades Kotlin as well as some other libraries to Kotlin 2.0.10. This upgrade is required due to side effects of iOS 18 and Xcode 16.

- ktor to 2.3.12
- kotlinx.serialisation to 1.7.1 (transitive, see https://github.com/ACINQ/lightning-kmp/pull/701)
- kotlinx.coroutine to 1.9.0 (transitive)
- kermit to 2.0.4 (transitive)
- skie to 0.8.4
- sqldelight to 2.0.2

# Issues found

### iOS: `SKIE` does not support kotlin 2.0.20 yet

We have to use Kotlin 2.0.10 instead, although `lightning-kmp` can keep using 2.0.20 without issues.

### `kotlinx.serialisation` v1.7.2 is not compatible 

This is a side effect of SKIE. `kotlinx.serialisation` v1.7.2 requires Kotlin 2.0.20 (because they use the new `kotlin.uuid.Uuid` object). In the meantime, we must use `kotlinx.serialisation` v1.7.1.

### iOS: Framework script removed 

In the build phases of `phoenix-ios-framework` we have a script that manually copied the `phoenix-shared` framework (and DSYM) generated by gradle in `phoenix-shared/build` into the Xcode `DerivedData` folder for the `phoenix-ios` application. 

This script now fails with Kotlin 2.0 because the files are already copied during the build process. @robbiehanson It should be investigated further but it seems this script is not needed anymore, and thus must be removed.

### iOS: Odd cache error

⚠️ **needs investigation**

An odd build error occurs in `:phoenix-shared:linkDebugFrameworkIosArm64`:

`
/Users/xxx/.gradle/caches/modules-2/files-2.1/app.cash.sqldelight/coroutines-extensions-iosarm64/2.0.2/51f8ddc76a4aa3d45402f46017e7b8405a5f6761/coroutines-extensions is cached (in /Users/xxx/.konan/kotlin-native-prebuilt-macos-x86_64-2.0.10/klib/cache/ios_arm64-gSTATIC-pl/app.cash.sqldelight:sqldelight-coroutines-extensions/unspecified/1v8rzgt0jt9l5.30bg0yuy4wun9/app.cash.sqldelight:sqldelight-coroutines-extensions-cache/bin/libapp.cash.sqldelight:sqldelight-coroutines-extensions-cache.a), but its dependency isn't: /Users/xxx/.konan/kotlin-native-prebuilt-macos-x86_64-2.0.10/klib/common/stdlib
`

Although it's mentioned in the log, it does not seem linked to the sqldelight coroutine extension. This error is "fixed" by removing the `kotlinOptions.freeCompilerArgs += "-Xallocator=std"` in `phoenix-shared/build.gradle.kts`. 

A similar report can be found [here](https://slack-chats.kotlinlang.org/t/18860817/trying-to-update-to-kotlin-2-0-0-and-when-i-build-my-ios-app).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Build
 build/
-.kotlin/
+.cxx
+.kotlin
 
 # Idea
 .idea
@@ -25,8 +26,6 @@ sdk-dependencies
 apk
 phoenix-legacy/release
 phoenix-legacy/debug
-.cxx
 
 # LangTool
 .langtool/deepl.authtoken
-
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -3,7 +3,7 @@ object Versions {
     const val secp256k1 = "0.14.0"
     const val torMobile = "0.2.0"
 
-    const val kotlin = "2.0.10"
+    const val kotlin = "2.0.21"
 
     const val ktor = "2.3.12"
     const val sqlDelight = "2.0.2"
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -3,10 +3,10 @@ object Versions {
     const val secp256k1 = "0.14.0"
     const val torMobile = "0.2.0"
 
-    const val kotlin = "1.9.22"
+    const val kotlin = "2.0.10"
 
-    const val ktor = "2.3.7"
-    const val sqlDelight = "2.0.1"
+    const val ktor = "2.3.12"
+    const val sqlDelight = "2.0.2"
 
     const val slf4j = "1.7.30"
     const val junit = "4.13"
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -1,5 +1,5 @@
 object Versions {
-    const val lightningKmp = "1.8.4"
+    const val lightningKmp = "1.8.5-SNAPSHOT"
     const val secp256k1 = "0.14.0"
     const val torMobile = "0.2.0"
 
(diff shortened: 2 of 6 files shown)


robbiehanson:
> it seems this script is not needed anymore, and thus must be removed.

I can confirm. It seems the bug was fixed in Kotlin, so the workaround is no longer needed.
However I did discover a new problem:

Screenshot 2024-09-21 at 3 43 30 PM

It seems the Framework is being copied within itself. So duplicated. Which means 2 copies of a the ~70Mb binary. So I added another build script to remove the duplicate (if found).

robbiehanson:
> This error is "fixed" by removing the `kotlinOptions.freeCompilerArgs += "-Xallocator=std"` in `phoenix-shared/build.gradle.kts`.

Unfortunately this breaks background payments on iOS: https://github.com/ACINQ/phoenix/pull/324

It's better than it was before, but the memory limit is still 24MB (for iOS 17), and I'm able to hit that pretty easily with 1 or 2 background payments.

dpad85:
Notes:
- the SKIE release v0.9.0 is now compatible with kotlin 2.0.20.
- we should move to the "direct integration" model : https://kotlinlang.org/docs/multiplatform-direct-integration.html. A test commit is available here: 41c1d151fc37fae662d2e91db614b691228d91a7

**Also**:
Commit 736d164 triggers a new build error (in `:phoenix-shared:linkDebugFrameworkIosArm64`):
```
e: java.lang.IllegalStateException: IrClassPublicSymbolImpl for kotlin/Function6|null[0] is already bound: 
CLASS IR_EXTERNAL_DECLARATION_STUB INTERFACE name:Function6 modality:ABSTRACT visibility:public superTypes:[kotlin.Function<R of kotlin.Function6>]
...
```
That error is due to the `kotlin.native.cacheKind=none` configuration and is seemingly caused by the tor-mobile-kmp library.

Removing that cache configuration triggers another error, `x is cached, but its dependency isn't`, which is caused by `-Xallocator=std`.

dpad85:
Superseded by https://github.com/ACINQ/phoenix/pull/654.

PR #629, opened 2024-09-20
(ios) Redesigned Send screen

robbiehanson:
Bolt 12, BIP-353 & Contacts create the need for a newly designed Send screen:

Notes:
- if the user starts typing, a search of their contacts will occur automatically
- if the user is typing a lightning address, then we'll automatically make smart suggestions to help auto-fill the domain portion
- we have a hard-coded list of well-known domain names
- in the future, contacts will also support lightning addresses. So in the future, the set of domain names will be (hard-coded list + any domains associated with existing contacts)

New users won't have any contacts. So we also have placeholders, so users understand what's possible:

Technical notes:

On iOS, we're no longer using the MVI pattern for the Send screen. In it's place is a new file `SendManager.kt`, which is basically a copy-paste of `ScanController.kt`, but modified to be stateless.


phoenix-ios/phoenix-ios/views/send/SendView.swift
```diff
@@ -11,338 +8,858 @@ fileprivate var log = LoggerFactory.shared.logger(filename, .trace)
...
-	@StateObject var mvi: MVIState<Scan.Model, Scan.Intent>
+	@State var inputFieldText: String = ""
 	
-	@Environment(\.controllerFactory) var factoryEnv
-	var factory: ControllerFactory { return factoryEnv }
+	@State var imagePickerResult: PickerResult? = nil
+	
+	@State var autocompleteSuggestions: [String] = []
+	
+	@State var sortedContacts: [ContactInfo] = []
```
dpad85, line 37: I have a few questions about the list of contacts displayed (especially when the SendView is initialised)
- is this `sortedContacts` ordered by latest paid ones?
  - if it is, is the code shared or specific to the iOS app?
- do you display all the contacts, or just a subset (e.g., the last 5 contacts)?
  - if only a subset, how can the user access the list of all contacts ?

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/SendManager.kt
```diff
@@ -0,0 +1,619 @@
...
+                val auth: LnurlAuth
+            ): Lnurl()
+        }
+    }
+
+    suspend fun parse(
+        request: String,
+        progress: (p: ParseProgress) -> Unit
+    ): ParseResult {
+
```
dpad85, line 143: We should find a way to make this method a bit smarter, and fail early whenever possible.

For example: if I enter `lnbc1xxxxxx`, we should know that it's a BOLT11 invoice because it starts like one, and if the BOLT11 parsing fails, return an error immediately instead of trying everything.

Another example: if I enter something containing an `@`, it's obviously an LN address, no need to try the rest. 

And if this address does not work, we should return a nice LN address error type, so that we can display an helpful message. In many cases, we end up returning a `BadRequest(..., reason = BadRequestReason.UnknownFormat)` error, which is unhelpful and a bad UX.

I'm not sure how much work it entails. If possible, we should seize the opportunity of this PR to fix this issue. Or do you think it should be done in another PR? It would also require specific tests.

dpad85 (commented):
I'm not able to test this branch, but have you checked if deep linking an URI still works?

> On iOS, we're no longer using the MVI pattern for the Send screen.

Does this change lead to a noticeable life improvement for developing on iOS? That's what we have been assuming, but it would be nice to confirm that. Maybe it's not as significant as we thought. It does look like you've been able to remove some code so it sounds good.

robbiehanson:
> I'm not able to test this branch

Understood. I'll rebase after #628 is merged.

> have you checked if deep linking an URI still works

Yup, this still works. It's actually better now. For example, say you launch the app via a URL like "phoenix://foobar@phoenixwallet.me":

- in the past, the user would only see the Home screen, and there was no visual indicator that the app was processing the request until the MVI layer completed, and we jumped to the Send screen
- now the Home screen shows the activity indicator that says "resolving lightning address", and even includes a cancel button

There are still some navigation issues on iOS 16. For example, if the app is open, and you're somewhere in Settings, then opening a link won't properly pop you from your current location to the Send screen. But this is mostly because navigation on iOS 16 is complicated and fragile, and is luckily now fixed on 17+.

> Does this change lead to a noticeable life improvement for developing on iOS ?

I would say Yes, it makes life easier. The pattern we've seen over and over again has been:

- we started with the MVI pattern when the app was super simple
- then we added feature X, Y and Z
- state management became much more complicated, and was better handled in the UI layer
- the MVI layer became more of a nuisance than a help
- then we dropped MVI in favor of simple StateFlow's and functions

Prior to this change we had some payment flows that were using MVI, and others that weren't (like Bolt12). Now everything has the same flow and structure, and it's easier to deal with in the code base.

phoenix-ios/phoenix-ios/views/send/SendView.swift
robbiehanson: The `sortedContacts` is the list of all contacts, sorted by name, as delivered by the database query.

The final plan is to have 2 different sections in the list:
- Recent Contacts
- All Contacts

Where "recent contacts" is maybe the 5 most recent that we've sent payments to.

Note that displaying "all contacts" here doesn't require any additional work. The list of contacts is already available in memory, and the UI `List` is "lazy", meaning it only bothers rendering what is visible on the screen.

Also, we can follow Apple's lead: in the Contacts app they display the alphabet on the trailing edge of the screen, making it super easy to jump to a letter, and find a certain person without typing.

However, we need to be smart about these features. For example, if the user only has a total of 5 contacts, it doesn't make sense to display the "recent contacts" section - only the "all contacts" section. And Apple doesn't bother to display the alphabet until the user has more than X total contacts. We should probably do something similar.

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/SendManager.kt
robbiehanson: That's a good idea. But probably best to do it in a separate PR. This was really just supposed to be a UI change, but turned into more because we dropped the MVI pattern. But at least, in terms of the code changes for this PR, the new SendManager is largely a copy-paste of pre-existing ScanController. But if we start changing a bunch of logic in the SendManager...

phoenix-ios/phoenix-ios/views/send/ParseResultHelper.swift
```diff
@@ -0,0 +1,125 @@
...
+		case let serviceError as SendManager.BadRequestReason_ServiceError:
+			
+			let remoteFailure: LnurlError.RemoteFailure = serviceError.error
+			let origin = remoteFailure.origin
+			
+			let isLightningAddress = serviceError.url.description.contains("/.well-known/lnurlp/")
+			let lightningAddressErrorMessage = String(
+				localized: "The service (\(origin)) doesn't support Lightning addresses, or doesn't know this user",
+				comment: "Error message - scanning lightning invoice"
+			)
```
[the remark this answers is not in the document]
robbiehanson: Done in bd9b069c

I moved all the iOS-specific error processing to the shared layer. Note this also includes the `IsWebsite` error.

Several times I've been at a restaurante/cafe, and they have a QR code to "pay with bitcoin". Except the QR code is just a website link, and the website has all the payment stuff. In the past, if you scanned the QR code with Phoenix then you just get a generic error. So on iOS, we had logic to detect this situation, and then we displayed a specific popover to the user, allowing them to copy the link, or open it with their browser.

phoenix-ios/phoenix-ios/views/contacts/ContactsList.swift
```diff
@@ -129,10 +133,34 @@ struct ContactsList: View {
 	}
 	
 	@ViewBuilder
-	func zeroMatches() -> some View {
+	func zeroMatchesRow() -> some View {
 		
-		HStack(alignment: VerticalAlignment.center, spacing: 0) {
-			Text("No matches for search").foregroundStyle(.secondary)
+		HStack(alignment: VerticalAlignment.center, spacing: 8) {
+			Image(systemName: "person.crop.circle.badge.questionmark")
```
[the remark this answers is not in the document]
robbiehanson: Good catch. Fixed in d7265396

phoenix-ios/phoenix-ios/views/send/SendView.swift
dpad85: Having 2 sections makes sense. For now, a list of all contacts is fine, especially if the list can be filtered in the text input.

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/SendManager.kt
dpad85: Alright, let's keep things focused.

PR #630, opened 2024-09-25
Support new on-the-fly funding

dpad85:
This PR adds support for the new on-the-fly funding based on liquidity ads.

*This description is a WIP.*

For the user, it's an UI change : the cost that is sometimes incurred when receiving payments is not displayed as fees anymore ; instead it's now displayed as a separate outgoing payment (but the cost stays the same).

This change will allow for more options regarding liquidity and funding in the future. For example, opportunistically request additional funding on-the-fly. 

More details about the underlying change can be found here:
- (bolt specifications) https://github.com/lightning/blips/pull/36
- https://github.com/ACINQ/lightning-kmp/pull/649
- https://github.com/ACINQ/lightning-kmp/pull/660


dpad85:
Superseded by #632 which does not need the Kotlin 2.0 update.

PR #633, opened 2024-10-04
(android) Add UI to spend from final wallet

dpad85:
Thanks to https://github.com/ACINQ/lightning-kmp/pull/700 (lightning-kmp v1.8.0), we can add the UI to spend confirmed funds in the final wallet from Phoenix (instead of having to use a 3rd party tool).

This PR adds a new UI similar to the one for spending cancelled swap-ins. The feerate can be customised (with a warning if it's too low). A notification has been added in the home screen to let the user know when there are funds in the final wallet (easy to miss otherwise).


dpad85:
image

image

---

image

image

image

image

PR #635, opened 2024-10-08
Ignore empty liquidity purchases

dpad85:
In some scenarios, we may have to execute liquidity purchases that request no liquidity (for example, channel creation from a swap-in). Due to limitations in how liquidity ads works, these purchases must request a minimal liquidity amount (we use 1 sat). This means that the UI displays 1 sat liquidity purchases. Moreover the fee incurred by these swap-ins is displayed twice (once in the liquidity purchase, and once in the swap-in details).

To avoid any confusion, this PR updates the aggregation SQL queries to ignore empty liquidity purchases (includes the CSV export).


pm47 (dismissed):

dpad85:
Superseded with https://github.com/ACINQ/phoenix/pull/636

Ignoring empty liquidity purchases will causes issues down the line. Instead these purchases are displayed with an improved UI.

PR #646, opened 2024-10-31
(ios) Add UI to spend from final wallet

robbiehanson:
This is the iOS version of #633 

&nbsp;&nbsp;&nbsp;&nbsp;

&nbsp;&nbsp;&nbsp;

On iOS, we re-used the same UI screen we were using for spending expired swap-ins. So just a little bit of refactoring was needed.


phoenix-shared/src/iosMain/kotlin/fr/acinq/phoenix/utils/LightningExposure.kt
```diff
@@ -417,3 +424,11 @@ suspend fun Peer.betterPayOffer(
     send(PayOffer(paymentId, payerKey, payerNote, amount, offer, fetchInvoiceTimeoutInSeconds.seconds))
     return res.await()
 }
+
+fun fakeWallet(amount: Satoshi, swapInParams: SwapInParams): WalletState.WalletWithConfirmations {
```
dpad85, line 428: This does not seem to be used anywhere, so it can be removed, or moved to the test module if needed.

dpad85 (approved):

PR #647, opened 2024-11-07
(android) Fix settings text of experimental features

Crazy-Unicooorn:
## Context of the issue
I've noticed that in the settings page, the "Experimental features" button was NOT translated even though upon clicking it, the settings title was shown translated (it is the same text).
Below are 2 screenshots. The first one shows the untranslated text, the second one shows the same text but available in French.
tempFileForShare_20241107-162325
tempFileForShare_20241107-162344

So I figured the stringResource() might be missing, and it seems to be what's happening.
So here's a simple attempt at fixing it.

## Disclaimer
I'm very new to this... both Git and AndroidStudio.
I've NOT tested my commit
Please review accordingly 😅


dpad85 (approved):
Good catch! We missed that one.

PR #648, opened 2024-11-07
(android) Improve French translation

Crazy-Unicooorn:
I have added some strings in values-fr 🇫🇷 
I'm a native french speaker
I have NOT tested my commit because I'm lazy and bad at figuring out how to build 😆 
Please review accordingly, thanks for your work guys!


dpad85 (approved):
Merci pour ce travail 😃 

> I have NOT tested my commit because I'm lazy and bad at figuring out how to build

No worries! Building the app is going to be much easier, especially on Android, when we remove the legacy module and its old dependencies that have to be built manually (eclair-core & a tor library).

PR #652, opened 2024-11-15
Prevent pending payments on same payment hash

dpad85:
This PR adds an additional control when paying a bolt11 invoice. We now not only check that the payment has not already been paid, but also that it's not pending either. Trying to pay the same payment hash in parallel can cause issues and channels being force-closed.

A new `BadRequestReason.PaymentPending` error type has been added.


phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/SendManager.kt
```diff
@@ -209,10 +211,16 @@ class SendManager(
         }
 
         val db = databaseManager.databases.filterNotNull().first()
-        return if (db.payments.listLightningOutgoingPayments(invoice.paymentHash).any { it.status is LightningOutgoingPayment.Status.Completed.Succeeded }) {
-            BadRequestReason.AlreadyPaidInvoice
-        } else {
-            null
+        val similarPayments = db.payments.listLightningOutgoingPayments(invoice.paymentHash)
+        // we MUST raise an error if this payment hash has already paid, or is being paid.
```
dpad85, line 215:
```suggestion
        // we MUST raise an error if this payment hash has already been paid, or is being paid.
```

PR #654, opened 2024-11-18
Upgrade to Kotlin 2.1.10

dpad85:
This PR upgrades the project to Kotlin 2. It supersedes https://github.com/ACINQ/phoenix/pull/628.

- kotlin updated to v2.1.10
- SKIE updated to v0.10.1

We also switch to the [direct integration technique](https://kotlinlang.org/docs/multiplatform-direct-integration.html) which is the standard and also fixes a build issue on some Xcode versions.


dpad85:
Note, to manage dependency versions we should also switch to the toml version catalog instead of the `buildSrc` technique. See https://github.com/ACINQ/lightning-kmp/pull/727.

phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/CurrencyManager.kt
```diff
@@ -647,7 +647,6 @@ class CurrencyManager(
         //
         val http = HttpClient {
             engine {
-                threadsCount = 1
```
dpad85, line 650: @robbiehanson This option has been deprecated at the error level and should not be used anymore, they recommend to use `dispatcher` instead. But the whole issue may have been fixed altogether in a previous version ? This needs some investigation.

robbiehanson: Yes, it looks like the original issue was fixed in Ktor v2.0.2. From the [release notes](https://github.com/ktorio/ktor/releases/tag/2.0.2):

> "Darwin engine: Client connection is closed after each request ([KTOR-4145](https://youtrack.jetbrains.com/issue/KTOR-4145))"

Which was a similar issue the what I reported in [KTOR-3362](https://youtrack.jetbrains.com/issue/KTOR-3362/IosClientEngine-should-use-a-shared-NSURLSession)

Also, in investigating this I discovered that api.coindesk.com appears to be completely shutdown. As in, I don't even see a DNS entry for it anymore. I'll create a new ticket for it.

robbiehanson (approved):

PR #662, opened 2024-12-12
Remove Tor embedded library and use onion addresses

dpad85:
This PR removes the [tor-mobile-kmp](https://github.com/ACINQ/tor-mobile-kmp) library embedded inside Phoenix. Access to Tor is not done by Phoenix anymore, instead, when Tor is enabled in the application, Phoenix will exclusively use onion addresses when connecting to the Lightning peer and to Electrum.

Access to the Tor network is delegated to a Tor proxy application that needs to be installed by the user (such as [Orbot](https://github.com/ACINQ/tor-mobile-kmp)).

This provides several improvements:
- Receiving payments in the background with Tor enabled will be more reliable, because the Tor proxy app will be able to run continuously (since it's a VPN application enjoying specific privileges).
- Using onion addresses ensures that the app cannot connect if Tor is unavailable for some reason.
- The previous Tor option only covered Lightning and Electrum connections, but not HTTP ones (which include API calls for feerates or btc/fiat rates).

Besides, the `tor-mobile-kmp` library was difficult to maintain. For example, it did not contain a recent Tor binary. Orbot will do a much better job at that. Our library also caused issues with the Kotlin 2.0 update, and with M3 chips.


robbiehanson:
iOS screenshots:

New configuration screens:







Notifying users of invalid Electrum configuration:





Modifying Electrum configuration:





When upgrading to the new version, **IF Tor is enabled**, we display this notice to the user:

dpad85:
Note that (on Android at least) we need to let users access the Settings screen from the initialisation screen, so they can activate Tor/set Electrum before they actually create or restore their wallet. But we'll do that in another PR, since this one is blocking other important PRs.

05nelsonm:
This breaks my heart.

I maintain the [kmp-tor](https://github.com/05nelsonm/kmp-tor) library and would love to see this feature reintegrated into Phoenix. Orbot has a plethora of issues and is unreliable.

dpad85:
There are indeed other good options which could replace our library (although note that Arti is officially not ready for production). But embedding Tor directly in the app inherently has drawbacks.

The issue is that Lightning is an interactive protocol. You need to respond to payment messages, and do so quickly. 

But when the app is in the background or stopped (which is often), the OS gives you a limited budget to process notifications in a timely manner (and they tend to shrink that budget in each new version). Having to first start Tor and connect before processing a payment is too much; it means that in most cases, background payments will fail.

Delegating the Tor connection to Orbot not only removes "work" from the Phoenix app, it is also persistent and more reliable because it is treated as a VPN connection.

CaverneCrypto:
I totally agree with both @05nelsonm and @MasterixCZ . This is very not UX friendly. Users do not want to switch app to enable Orbot each time they want to process a payment. 

Yes, the previous implementation was long to start and to connect to tor but I worked (without any click of the user)!

In my opinion, remove tor from phoenix is not going in the right direction for the privacy of user, because they will now just not enable tor on their wallet...

I already have some complain about this from friend who use Phoenix often.

PR #667, opened 2025-01-09
(ios) Display FinalWallet balance on HomeScreen

robbiehanson:
If the user has funds in their FinalWallet, that information is now included in the "incoming balance" information on the HomeScreen. Also, tapping that balance will now show a popover that mirrors recent changes on Android:

Also added a badge for the HomeScreen that counts in-flight payments (mirrors Android UI changes in #522 ):


dpad85:
Note, the `pendingChannelsBalance` coming from the `BalanceManager` is going to be removed in the database migration update. It's not useful anymore, it's always 0 since the new on-the-fly funding mechanism (https://github.com/ACINQ/lightning-kmp/pull/649) which circumvents the issue altogether. So we'll soon be able to clean up some code.

dpad85:
Nitpick, I think the popover would look better if the background was gray and the buttons white.

robbiehanson:
Updated colors:

PR #671, opened 2025-02-07
Remove the phoenix-legacy module

dpad85:
This PR removes the old legacy Android application, and the code in the modern app that was related to the legacy app.

For context, the legacy app, contained in this `phoenix-legacy` module, is first version of Phoenix release in 2020, build around the `eclair-core` Scala library. It's been deprecated since Phoenix Android v2.0, released in August 2023. 

Since October 2024, all channels that have ever been created with the legacy app have been closed, and the legacy application is now unusable, and it's now time to remove all the code.

Note that the custom TLV for Phoenix migration that was added [in lightning-kmp](https://github.com/ACINQ/lightning-kmp/pull/319) and eclair can now be removed.

### Legacy channels recovery

Legacy channels have all been closed. However some of these channels that were created in specific versions may need to be manually recovered. The easiest way to do that for users is to use the recovery tool present in the legacy app. This tool will not be available after this PR. but users can still download Phoenix v2.4.5 (or versions prior) and perform the recovery there.


dpad85:
This PR builds upon https://github.com/ACINQ/phoenix/pull/662 and will be rebased.

PR #673, opened 2025-02-18
(android) Improve scanner performance with CameraX

dpad85:
CameraX brings many compatiblity fixes and should help make the scanner work better on all the various Android devices that exist in the wild. It also provides an easy and efficient way to perform image analysis, which lets us directly use and configure the zxing QRcode reader.

Besides, the camera starts faster and we are now able to zoom and tap to focus the camera.

Note that this update would also make it very easy to switch to another QR code scanning lib, like [Google's MLKit](https://developers.google.com/ml-kit/vision/barcode-scanning) if performance issues with the scanner remain. However that library is closed source, so Zxing is preferred.


dpad85:
Also added a few UI fixes that spawned after the compose upgrade in the kotlin 2 pull request.

PR #675, opened 2025-02-24
Add more details to refund signing result

t-bast:
We seem to sometimes obtain invalid sigs from Phoenix users. Having more details should help troubleshoot the issue.

Note that I wasn't able to build Phoenix locally (IntelliJ issues) to verify that the code correctly compiles :/

### DIFF
--- a/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/utils/channels/SpendChannelAddressHelper.kt
+++ b/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/utils/channels/SpendChannelAddressHelper.kt
@@ -25,6 +25,8 @@ import fr.acinq.bitcoin.SigHash
 import fr.acinq.bitcoin.SigVersion
 import fr.acinq.bitcoin.Script
 import fr.acinq.bitcoin.Transaction
+import fr.acinq.bitcoin.TxId
+import fr.acinq.bitcoin.byteVector
 import fr.acinq.lightning.channel.states.ChannelStateWithCommitments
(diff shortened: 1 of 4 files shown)


phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/utils/channels/SpendChannelAddressHelper.kt
```diff
@@ -112,8 +115,12 @@ object SpendChannelAddressHelper {
             val fundingScript = Scripts.multiSig2of2(localFundingKey.publicKey(), pubkey)
 
             val sig = Transactions.sign(tx = tx, inputIndex = 0, Script.write(fundingScript), amount, localFundingKey)
-            return SpendChannelAddressResult.Success(localFundingKey.publicKey(), sig)
-
+            val signedData = tx.hashForSigning(0, Script.write(fundingScript), SigHash.SIGHASH_ALL, amount, SigVersion.SIGVERSION_WITNESS_V0)
```
pm47, line 118: We should probably log the `fundingScript`, `amount` and `signedData` here.

pm47: > we know the amount because the user copy-pasted it

Sometimes users behave in unexpected ways 😅

dpad85 (commented):
I think it's good to go, it would be nice to have this improvement in the next version.

pm47 (approved):

PR #677, opened 2025-02-25
Clarify UI for automated liquidity purchases

dpad85:
With the database rework, we thought that we would be able to simply filter the `AutomaticLiquidityPurchasePayment` payments from the payments history. 

However it has some side-effects on the UI because of the way we build that payments history (the total number of payments in the db does not match the number of payments in the list displayed in the UI).

Thus we should keep displaying these `AutomaticLiquidityPurchasePayment`s. 

However, some adjustments are needed to avoid confusion: the fee of these purchase is already taken into account in the amount of the incoming payment that triggered it. See the example below, the amount received before fees was 400k. This is confusing and it feels like we paid twice.

So we now hide the amount of the channel management row. 
This is done by checking `payment.incomingPaymentReceivedAt` : if `null`, we should show the amount ; if not, it means the related payment was received and so we should hide the channel management amount.

The details screen also contains a button to navigate to the relevant incoming payment :

@robbiehanson You can use `PaymentsManager.getIncomingPaymentForTxId` to get that incoming payment from the auto purchase details screen

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-lightningkmp = "1.8.5-SNAPSHOT"
+lightningkmp = "1.9.0"
 secp256k1 = "0.17.1"
 
 kotlin = "2.1.10"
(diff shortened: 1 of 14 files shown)


dpad85:
Also added a commit to improve reconnection on Android.

PR #685, opened 2025-03-12
(android) Add manual payments import/export

dpad85:
This PR adds a system to manually import/export the wallet's payments database. It allows users to transfer their payments history from one device to another. This process is pretty basic but could be automated.

### Exporting payments

A new section for payments database export has been added below the CSV export. The user uses the device file browser to pick the location of the export, for example the Downloads folder.

image

### Importing payments

When restoring a wallet, the user can also pick a payments backup file after typing down the seed. This is optional. If the user does not import a payments backup or if the file picked is invalid, it's not blocking, the wallet can still be restored.

image

image

image

### Permissions

The app does not need any new permissions because access to the file system is done through the user manual input. If we want to automate the export in a background job, we'll need to persist the access granted by the system to the export file.

### Encryption

The file is zipped and encrypted with ChaCha20Poly1305, using a custom derivation key path. We use a version byte in case of future breaking changes. @sstone I've also added an alternative with AES and javax crypto for comparison. See `EncryptedData`.

This PR supersedes #561 which did too many things at once. It is almost ready, but we still need to segregate the encryption key with the chain to avoid issues, and add tests.


phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/security/EncryptedData.kt
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * This object represents data encrypted with a key derived from the wallet seed. The type of encryption and key used depends on [version].
+ *
+ * Note that the data is zipped before encryption.
+ *
+ * @param version the version used for the data encryption and for serializing this object
+ * @param data a byte array containing the zipped & encrypted payload, followed by the stuff needed for decryption (which depends on the version)
+ */
+class EncryptedData(val version: Version, val data: ByteArray) {
```
sstone, line 43: It looks that the use case here is very similar to the encrypted channels backup extensions we added to some LN messages. We should probably export encryption methods from `lightning-kmp` instead of duplicating the code (and it would take care of private key handling) ?

phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/security/EncryptedData.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    sealed class Version(val code: Byte) {
+        // use ChaCha20Poly1305, inspired from the channel backup encryption code in lightning-kmp
+        // see https://github.com/ACINQ/lightning-kmp/blob/feda82c853660a792b911be518367a228ed6e0ee/modules/core/src/commonMain/kotlin/fr/acinq/lightning/serialization/channel/Encryption.kt#L14
+        data object V1 : Version(1) {
+            fun getKey(keyManager: LocalKeyManager) = keyManager.privateKey(KeyPath("m/150'/1'/0'")).value.toByteArray()
+        }
+        // use AES with javax.crypto
+        data object V1_BIS : Version(127) {
+            val algorithm = "AES/CBC/PKCS5PADDING"
```
sstone, line 53: We should just use `ChaCha20Poly1305`.

phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/security/EncryptedData.kt
dpad85: Yes, it is heavily inspired by that code, but I don't think it should be moved to lightning-kmp in a generic encryption helper:
- the key used is specific to Phoenix (and in fact we now use the `cloudKey` key already used by the iOS app for cloud backups), so it should stay in the Phoenix repo
- Most of the code in `EncryptedData` is related to serialisation and zipping the data. The code specific to encryption/decryption is actually very small. Really it's only how we compute the nonce, and then calling `encrypt/decrypt`. We would not save much in Phoenix, and it would pollute lightning-kmp.

robbiehanson:
Looks good. This is a good first step. Ultimately most users will want this process automated. And it looks like there's a path to get there using this foundation.

PR #690, opened 2025-04-01
(ios) UI issues regarding received payments via Bolt 12

robbiehanson:
In the shared kotlin layer, `PaymentsManager.lastCompletedPayment` was changed at some point to **exclude** incoming Bolt12 payments. This resulted in the following bugs in iOS:

1. Normally, when an incoming payment is received, the iOS app displays a sheet (that slides up from the bottom of the screen) alerting the user to the payment. This wasn't occurring for incoming Bolt12 payments.
2. When inside the "Receive" screen, and looking at the Bolt12 info, an incoming Bolt12 payment should dismiss the screen, and show the user the Home screen's sheet. This wasn't occurring.
3. When a Bolt12 payment is received while the app is in the background, it should display a notification about the received payment. Instead it was displaying a notification of "missed incoming payment" even though the payment was received.

Problem 3 could also result in the database not properly updating in the foreground app, and the UI wouldn't display the received payment.

This PR changes `lastCompletedPayment` to include **ALL** lightning payments.
This PR addresses ticket [84169632](https://phoenix.jitbit.com/Ticket/84169632)

In iOS we create a derivative flow for sub-categories. I.E. `lastCompletedIncomingPayment` is a derivative flow that only emits incoming payments. If we need a derivative flow that excludes incoming Bolt12 for Android, we could create a similar "sub-flow".

---
Note that changing `lastCompletedPayment` affects 2 files in phoenix-android:

AppView.kt:
```kotlin
val lastCompletedPayment by business.paymentsManager.lastCompletedPayment.collectAsState()
lastCompletedPayment?.let { payment ->
    LaunchedEffect(key1 = payment.id) {
        navigateToPaymentDetails(navController, id = payment.id, isFromEvent = true)
    }
}
```

ReceiveLightningView.kt
```kotlin
// refresh LN invoice when it has been paid
val paymentsManager = business.paymentsManager
LaunchedEffect(key1 = Unit) {
    paymentsManager.lastCompletedPayment.collect { completedPayment ->
        if (state is LightningInvoiceState.Show && completedPayment is Bolt11IncomingPayment && state.invoice.paymentHash == completedPayment.paymentHash) {
            vm.generateInvoice(amount = customAmount, description = customDesc, expirySeconds = defaultExpiry)
            feeWarningDialogShownTimestamp = 0 // reset the dialog tracker to show it asap for a new invoice
        }
    }
}
```

### DIFF
--- a/phoenix-ios/phoenix-ios/views/receive/LightningDualView.swift
+++ b/phoenix-ios/phoenix-ios/views/receive/LightningDualView.swift
@@ -726,18 +726,43 @@ struct LightningDualView: View {
 	func lastIncomingPaymentChanged(_ lastIncomingPayment: Lightning_kmpIncomingPayment) {
 		log.trace("lastIncomingPaymentChanged()")
 		
-		guard let model = mvi.model as? Receive.Model_Generated else {
+		let state = lastIncomingPayment.state()
+		guard state == WalletPaymentState.successOffChain else {
+			log.debug("lastIncomingPaymentChanged(): state != .successOffChain")
 			return
 		}
 		
-		let state = lastIncomingPayment.state()
-		if state == WalletPaymentState.successOffChain {
+		guard let lightningPayment = lastIncomingPayment as? Lightning_kmpLightningIncomingPayment else {
+			log.debug("lastIncomingPaymentChanged(): not a Lightning_kmpLightningIncomingPayment")
+			return
+		}
+		
+		var didCompletePayment = false
+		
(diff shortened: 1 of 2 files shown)


dpad85 (approved):
> was changed at some point to exclude incoming Bolt12 payments

Good catch, that's my bad I've changed that in the db rework but there were side effects.

Note to self, fix the Android app to ignore bolt12 incoming payments events.

PR #697, opened 2025-04-16
(ios) Export BIP-321 URI via NFC (host card emulation)

robbiehanson:
https://github.com/user-attachments/assets/7213f43d-ffcf-426b-9e2b-c41d2986ea45

---

Technical notes:

- Reading an NFC tag is very simple (and doesn't require any special permissions from Apple)
- Performing HCE (host card emulation) and sending a tag is straight-forward once you have the building blocks in place (NDEF protocol stuff)
- HCE requires iOS 17.4 or later
- HCE will **only** work if your device is signed into iCloud, and your associated country is part of the European Economic Area
- For **development** builds, you can test with a device running iOS 18.2 or later (even if you're outside the EEA)
- Apple will need to accept our request to use HCE in our product
- The existing code we have in place for the SendManager is all we need to parse the received URL (no changes to kotlin shared layer needed).


dpad85 (commented):
Looks great!

dpad85 (approved):

PR #699, opened 2025-04-23
(android) Support Lightning addresses in contacts

dpad85:
This PR matches the changes made in the iOS app with #640 (support for lightning addresses within the Contacts functionality).

### Changes in `ContactInfo`

@robbiehanson For convenience I've modified `ContactInfo` to make `ContactOffer` and `ContactAddress` depend on a sealed class :
```kotlin
sealed class ContactPaymentCode {
    abstract val id: ByteVector32
    abstract val label: String?
    abstract val createdAt: Long
    abstract val paymentCode: String
}
```

This lets me factorise a lot of UI code in the Android app. 

### Difference with the iOS app

The offers and the Lightning addresses for a contact are displayed in the same list. The UI is more compact and easier to understand. See the starblocks screenshot above.

Also I've added a prominent "Send a payment" button in the contacts dialog. @robbiehanson wdyt? The payment code used is `ContactInfo.paymentCodes.firstOrNull`.


dpad85:
Another important thing happens in [8848fe5](https://github.com/ACINQ/phoenix/pull/699/commits/8848fe511fcdf9c4a2d4ee668b351811567a7734): the flows listing payments are now combined with the contacts list flow from `SqliteContactsDb`. 

This way, the payments pages are automatically refreshed when contacts change, which avoids having to manually observe that from the UI and manually refresh the payments history.

@robbiehanson I have not tested it on iOS, so it needs investigation to check if there are side effects or performance issues.

robbiehanson:
> The offers and the Lightning addresses for a contact are displayed in the same list. The UI is more compact

I'm a fan of this change. I will port this to iOS too.

> Also I've added a prominent "Send a payment" button in the contacts dialog. wdyt?

In iOS, there is also a "send payment" button, but it only appears when the contact has exactly 1 payment code. Otherwise, that button is hidden, and the user must pick a specific payment code to use.

> The payment code used is `ContactInfo.paymentCodes.firstOrNull`

I don't have a problem with this as long as the user can also choose the specific payment code from the list.

On iOS, I took some cues from how Apple handles contacts with multiple phone numbers. So if you open Phoenix, and go to Send, and select "Starblocks" (which has 2 payment codes), what does the UI do? On iOS, the list expands and shows the 2 payment codes, so the user can quickly tap which one they want.

robbiehanson:
> the flows listing payments are now combined with the contacts list flow

This is working well. I was able to remove a few hacks on the iOS side

robbiehanson (approved):

dpad85:
> I don't have a problem with this as long as the user can also choose the specific payment code from the list.
> 
> On iOS, I took some cues from how Apple handles contacts with multiple phone numbers. So if you open Phoenix, and go to Send, and select "Starblocks" (which has 2 payment codes), what does the UI do? On iOS, the list expands and shows the 2 payment codes, so the user can quickly tap which one they want.

It's not as good as on iOS, the UI does not let you pick a specific payment code, just the `paymentCodes.first`. I'll fix that.

dpad85:
> > I don't have a problem with this as long as the user can also choose the specific payment code from the list.
> > On iOS, I took some cues from how Apple handles contacts with multiple phone numbers. So if you open Phoenix, and go to Send, and select "Starblocks" (which has 2 payment codes), what does the UI do? On iOS, the list expands and shows the 2 payment codes, so the user can quickly tap which one they want.
> 
> It's not as good as on iOS, the UI does not let you pick a specific payment code, just the `paymentCodes.first`. I'll fix that.

Will fix in a future PR, with an UI similar to what's done on iOS.

PR #701, opened 2025-04-25
User-provided miner fee for simple close

robbiehanson:
UI changes to allow the user to choose the miner fee when draining the wallet (simple close):


dpad85:
Android UI:

PR #711, opened 2025-05-19
Add PIN code for spending

dpad85:
Spending operations (including swap refund or channel closing) can be protected behind a 6-digits PIN. This PIN code is different from the screen lock authentication. This allows sharing the wallet with someone else so that they can browse the app and receive payments, but not spend.

### Payment flow change

If this option is enabled, the "Pay" button triggering the spending event prompts for the PIN first.

This applies to the following operations:
- spending with Lightning (Bolt11, Bolt12, LNURL-pay)
- spending on-chain (splice-out, mutual close, force close, swap-in refund, final wallet)
- purchasing liquidity

In all cases the PIN prompt happens at the last moment. For example, with splice-out, the prompts shows up when the user taps the confirm button for the capacity disclaimer :

image

### Recovery phrase

Also affected is the recovery phrase display ; the phrase should not be visible if the user does not know the spending PIN. That part is still WIP.

### Access control settings change

There are now two sections in the Access control page: one for opening the app, and one for spending control. Creating the Spending PIN works the same as the Screen-lock PIN, the PIN is 6 digits and the user must confirm it before the option is enabled.

The wording is WIP, @robbiehanson if you have any suggestions don't hesitate.


dpad85:
Note, we should also add an option to scramble the PIN keyboard.

PR #712, opened 2025-05-20
Update README.md

xmready:
Changed download text links into icons and added Obtainium as a download source


xmready:
Screenshot From 2025-05-20 19-06-32

dpad85:
Sorry, we only list download links that we manage ourselves. The hash does help savvy users, but a lot of people will not check, and there's also no guarantee that the source is always reliable.

PR #717, opened 2025-05-24
ci: update actions to pinned SHA numbers

### LINKED ISSUE #716: ci: actions in workflow should be pinned to commit SHA not tags
It is best practice for actions in a workflow to be pinned to a specific commit SHA rather than using the latest tag. This is because using the latest tag opens an attack vector for malicious packages to be pulled in. Some examples include:

- Upstream action publishes a commit with malicious code in it
- Upstream action does not have rules set correctly, and a malicious actor is able to publish a new tag from a private branch
- Upstream action does not have GPG signing required, and a malicious actor is able to publish a new commit that appears like an automated commit (fake dependabot/renovatebot).

With actions pinned to a specific commit, we should have dependabot update the various packages on a weekly cadence and roll the SHA forward after review from the team.

andrewb1269hg:
**Description**:

Update the actions in the `testnet-build.yml` file to pinned commit SHAs.

The following actions were updated to the latest version:
`actions/checkout` previously `v2`, now `v4.2.2`
`actions/setup-java` previously `v1`, now `v3.14.1`

**Related Issue(s)**:

Fixes #716

### DIFF
--- a/.github/workflows/testnet-build.yml
+++ b/.github/workflows/testnet-build.yml
@@ -17,12 +17,12 @@ jobs:
     # 1 - checkout project and dependencies
 
     - name: Checkout Phoenix
-      uses: actions/checkout@v2
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         path: phoenix
 
     - name: Checkout lightning-kmp
-      uses: actions/checkout@v2
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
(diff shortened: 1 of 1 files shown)


dpad85 (changes requested):
Thanks for making this PR. The changes in the github action look good (though note that this action builds a testnet version of Phoenix, not the mainnet one).

About Dependabot : dependabot alerts are already enabled in the repository and I think that's sufficient. I don't think we need Dependabot to open Pull Requests as well. If there is a problem we'll fix it manually. Can you remove the `dependabot.yml` configuration file?

PR #721, opened 2025-05-28
Update README.md - package id and fingerprint

xmready:
There should be no objections with README improvement as it benefits users like myself who verify the package id and fingerprint.


xmready:
> we'd need to update it for every release and since it isn't signed, github (or any attacker) could change it without anyone noticing easily.

My understanding is this fingerprint is of the signing certificate and it does not change between releases, which is the exact reason apps like AppVerifier exist. Your concern about it nor being signed is valid.

> The information is already present in the Release page, alongside the APK

I don't see this fingerprint listed anywhere in the release page where a user can easily verifiy it against an app like AppVerifier.

Issue #639, opened 2024-10-11
Support selecting more than one fiat currency


Giszmo:
I often need to pay stuff in different currencies and it's cumbersome to prepare an on-chain transaction and lose the recipient while switching to another fiat currency.

Personally I couldn't care less about easily switching between the 4 bitcoin denominations but my screen height would allow for an additional 10 fiat currencies in the send screen currency dropdown.

As a concrete suggestion, please make the currency setting multi-choice and remember the last one used as default.

robbiehanson:
Dev notes:

On iOS this is handled thru the "currency converter" tool. See #247 for more info.

robbiehanson:
Fixed in PR #726

PR #726, opened 2025-06-27
(android) Support multiple fiat currencies

dpad85:
This PR matches the behaviour of the iOS app, where amount inputs can be entered in various currencies, to make it easier to use Phoenix when travelling.

There is still a primary Bitcoin and Fiat currency in the app settings, and they define how amounts are displayed throughout the app, just like before. 

However inputs now accept an amount in another fiat or bitcoin unit than the primary options. It means that if the user is using Euro usually and want to keep it that way, but travels to the US, he will be able to pick the USD currency when sending a payment, or when generating an invoice.

A currency converter has been added to facilitate converting amounts to other currencies. This converter is how the user can add new currencies.

Related issue: #639

## Default option

image

image

## Adding another currency

image

image

image


dpad85:
This PR is stable and will not change too much, so you start the translation now. Thanks!

PR #727, opened 2025-07-08
Updated Czech translation

MasterixCZ:
After a long time, I went through the Phoenix strings again and translated the new and missing ones. I also improved the already translated strings to make them make more sense given their context and I edited the important_strings.txt and strings.txt files in the values-cs folder so that their structure matches the original (it has diverged quite a bit over time)


phoenix-android/src/main/res/values-cs/strings.xml
```diff
@@ -751,20 +963,45 @@
...
+    <string name="nfc_button">NFC</string>
+    <string name="nfc_hce_emulation_title">Podržte telefon v blízkosti NFC čtečky</string>
+    <string name="nfc_reader_title">Připraveno na scan</string>
+    <string name="nfc_reader_desc">Podržte telefon v blízkosti NFC zařízení</string>
+
+    <!-- Currency converter -->
+
+    <string name="converter_title">Převodník Měn</string>
+    <string name="converter_done_button">Hotovo</string>
+    <string name="converter_placeholder_label">Vložte částku v %1$ | 919 satss</string>
```
dpad85, line 1001:
```suggestion
    <string name="converter_placeholder_label">Vložte částku v %1$s</string>
```

dpad85 (commented):
Thanks for this work. Just one nitpick at the end but it looks good.

Issue #743, opened 2025-09-12
Wallet stopped working for on-chain payments because of missing fee rate


Giszmo:
Whenever I try to make a payment on-chain, the wallet asks me to enter a valid fee rate and the fee rate update spinner keeps spinning for minutes and I think I cannot manually set a fee rate. This issue persists since a week now. I tried on maybe 5 occasions.

blatte:
I tried again and it finally worked...

dpad85:
It's likely due to mempool.space filtering some IPs (that's where the fee estimation comes from).

PR #745, opened 2025-09-18
Add recommended feerate flow in peer manager

dpad85:
This recommended feerate relies on mempool.space estimation for the half-hour target, with a fallback to the peer funding feerate if mempool.space is unavailable. This is more robust than just relying on mempool.space, and the peer funding feerate is much more pertinent than a static hard-coded fallback.


dpad85:
@robbiehanson Context is #743, which was an Android issue but may affect iOS as well, it's worth looking into.

PR #752, opened 2025-10-06
(ios) Original fiat values were missing due to recent changes

robbiehanson:
In PR #748, the `appConfigurationManager.preferredFiatCurrencies` was dropped and then re-added, and I missed this detail. Which meant the originalFiatValues were missing for all incoming/outgoing payments.


dpad85 (approved):
Good catch!

PR #767, opened 2025-10-30
(android) Fix per-wallet theme loading

dpad85:
This PR fixes an issue where the Android app would always use the system theme, and not the user theme. Phoenix now loads correctly the theme selected for a given wallet. Note that this theme can be different between wallets.

If no wallet has been selected yet (e.g. we're on the wallet selector screen), the app will default to the last used wallet's theme, or will fall back to the system theme.

This PR also fixes an issue where switching to another wallet from the settings menu would not load said wallet immediately, and would instead load the wallet selector screen, forcing the user to click on the wallet again.

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/utils/datastore/GlobalPrefs.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/utils/datastore/GlobalPrefs.kt
@@ -129,7 +129,7 @@ class GlobalPrefs(private val data: DataStore<Preferences>) {
         if (code == null) it.remove(SHOW_RELEASE_NOTES_SINCE) else it[SHOW_RELEASE_NOTES_SINCE] = code
     }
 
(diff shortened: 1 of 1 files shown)


phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/utils/datastore/GlobalPrefs.kt
```diff
@@ -125,6 +128,10 @@ class GlobalPrefs(private val data: DataStore<Preferences>) {
     suspend fun saveShowReleaseNoteSinceCode(code: Int?) = data.edit {
         if (code == null) it.remove(SHOW_RELEASE_NOTES_SINCE) else it[SHOW_RELEASE_NOTES_SINCE] = code
     }
+
+    /** Fallback theme used by the app when no specific is used and no user prefs is available. It should be set to the theme used by last active user. */
```
dpad85:
```suggestion
    /** Fallback theme used when no wallet theme is available yet. Set to the theme used by last active wallet. */
```

PR #774, opened 2025-11-04
Catch channel data deserialisation failures

dpad85:
Local channel data may not be readable, e.g. if the serialisation support for the format used has been dropped in lightning-kmp. This will crash the app. We should instead ignore local channels data when that happens (falling back to the peer backup), and log the error.


phoenix-shared/src/commonMain/sqldelight/channelsdb/fr/acinq/phoenix/db/sqldelight/ChannelsDatabase.sq
```diff
@@ -35,7 +35,7 @@ closeLocalChannel:
 UPDATE local_channels SET is_closed=1 WHERE channel_id=?;
 
 listLocalChannels:
-SELECT data FROM local_channels WHERE is_closed=0;
+SELECT channel_id, data FROM local_channels WHERE is_closed=0;
```
dpad85: Adding the `channel_id` here allows us to reuse in `listLocalChannels` the `mapChannelData` already used by `getChannel`. Also, lets us log the id of the faulty channel.

dpad85:
Superseded by #775.

PR #779, opened 2025-11-06
Improve electrum tests and remove a few servers

dpad85:
This PR improves the mainnet Electrum servers test (add a `getConfirmation` test). It also removes a few servers that do not perform well currently.


robbiehanson (approved):
Great idea !

PR #780, opened 2025-11-07
Use default dispatcher for phoenix-shared coroutine scopes

dpad85:
We've been used the `MainScope` until now for all our coroutines, a scope that relies on the UI dispatcher. This is not suited for many of the phoenix-shared coroutines which are doing background stuff and should not run on the UI thread. I believe it was done this way for historical reasons, back when kotlin multiplatform had issues with background operations on iOS, but that was fixed when we adopted the new memory model (see #299).

This PR defines a `defaultScope` method that uses the predefined `Default` dispatcher:

> uses a common pool of shared background threads (...) an appropriate choice for compute-intensive coroutines that consume CPU resources.
> https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/

There's also a `Dispatcher.IO` dispatcher that could be used for coroutines doing I/O operations.

Note: we've been already using `Dispatchers.Default` for database queries, but we did that by explicitly switching context with `withContext`. @robbiehanson I think that this PR could allow us to remove those switches and simplify code (and also ensure we're not doing db stuff on the Main thread).


robbiehanson (approved):
Looks good. And the iOS app feels more responsive now.

PR #785, opened 2025-12-23
(android) Fix German translation

real-or-random:
"Bewerbung" was simply wrong. It means "application" but in the sense of "job application".


dpad85 (approved):
Thank you for the fix!

PR #801, opened 2026-03-23
fix(id): shorten FAQ button label and use string resource for inbound liquidity

Rovikin:
- Shorten FAQ button label from 'Pertanyaan yang Sering Diajukan (FAQ)' to 'FAQ' to fit UI constraints
- Replace hardcoded 'Inbound liquidity' string in WalletView.kt with string resource `channelsview_inbound`


dpad85 (changes requested):
Thank you for the fix.

A few changes are needed:
- the `values-id` folder and its content need to be removed. Although `id` is the ISO code for Indonesia, Android uses `in` for the resource files instead (see commit https://github.com/ACINQ/phoenix/commit/3f3db8b8b949167ca745b5ee4b17de09cf85ce13)
- the `R.string.channelsview_inbound` entry for Indonesian needs to be added to `values-in/strings.xml`
- the `R.string.channelsview_inbound` entry for English must be added to `values/strings.xml` (that's mandatory, it will be the default value for other languages)

dpad85 (approved):

PR #812, opened 2026-05-26
(android) Block pointer events from falling through the screen-lock overlay

### LINKED ISSUE #733: UI elements accessible through lock screen when app returns from background
## Bug description
On Android, when the Phoenix wallet is locked (showing "Unlock to continue" screen), UI elements from the underlying payment screen apparently remain interactive and can be triggered by tapping on their expected positions, even though they're not visible. In this case, I could open the "Add a custom description to this payment" pop-up. What else could be exploited by this issue..?

## Steps to reproduce
1. Open Phoenix wallet and create an invoice
2. Share the invoice via another app, say a chat messaging app (this switches to another app)
3. Return to Phoenix wallet later. The app will be locked showing "Unlock to continue"
4. Tap in the middle area of the screen, slightly below the "Unlock to continue" text
5. The "Add a custom description to this payment" popup appears, even though the app is locked

## Environment
- **Phoenix version**: 2.6.2
- **Android version**: 15

## Screenshot
I marked the location where I tapped in the screenshot:

Image

PeterXMR:
Closes #733.

## Summary

When the wallet is locked and showing "Unlock to continue", pointer events in empty regions of the lock overlay can reach interactive elements drawn on the destination underneath, allowing those underlying controls to be triggered without unlocking the wallet. The original report describes opening the "Add a custom description to this payment" sheet through the lock screen.

## Change

Wrap the lock-prompt content in a Box whose `pointerInput` consumes any pointer event still unconsumed during the Final pass. Children with their own pointer handlers (the biometric / PIN code buttons inside the lock prompt) consume their events first during the Main pass, so the unlock controls remain interactive while empty-space taps no longer leak through.

[phoenix-android/.../StartupView.kt](https://github.com/ACINQ/phoenix/blob/master/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/startup/StartupView.kt) only — `LoadWallet`'s `isScreenLockRequired == true` branch. `contentAlignment = Alignment.Center` is set on the new Box so the snowflake icon and "Unlock to continue" text remain visually centered (matching master).

## Verification

Manually verified on a Pixel 4a API 34 emulator with a wallet that has Lock PIN enabled (1-minute auto-lock):

1. Navigate to **Receive → Lightning**.
2. Send the device to sleep via `KEYCODE_POWER` to trigger `ACTION_SCREEN_OFF` → `clearActiveWallet()` and re-enter the locked state.
3. Wake the device and dismiss the auto-popped PIN sheet (Back).
4. With "Unlock to continue" + the bottom PIN code button visible, tap at coordinates `(574,1522)`, `(354,1358)`, `(726,1358)` — these match where the underlying Lightning Receive screen's "Add amount and description" link, "Single use" tab, and "Reusable" tab live.

A/B comparison:

- **master**: the first tap caused the underlying screen's element to fire — the PIN sheet opened and subsequent taps registered as digit input (screenshot diff visible).
- **this PR**: all three taps produce byte-identical screenshots — no state change, no underlying element triggered.

Tap on the lock screen's own "PIN code" button still opens the PIN sheet, confirming that the legitimate interactive elements of the lock prompt remain reachable.

## Test plan

- [x] Build passes (`:phoenix-android:assembleDebug`)
- [x] Functional A/B on emulator: tap-through no longer triggers underlying screen
- [x] Lock screen's PIN code button still responds normally

### DIFF
--- a/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/startup/StartupView.kt
+++ b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/startup/StartupView.kt
@@ -51,6 +51,8 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.input.pointer.PointerEventPass
+import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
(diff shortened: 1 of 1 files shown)


dpad85:
Apologies for not reviewing this sooner. I'm closing this PR and #733 as the issue is not relevant anymore.

Issue #733 was fixed in 2025 with #742, which added multi-wallet support. Before that, the screen-lock was indeed a dialog layered over the UI, and it did not consume click events. But #742 changed that. Access control is now done via navigation logic, and there's no business UI underneath the screen-lock.

PR #814, opened 2026-06-02
(ios) Fix Send footer clipping with large Dynamic Type

Vidarte-Alberto:
## Problem

When the user has an accessibility font size enabled, the four action buttons in the Send screen footer (`paste`, `choose image`, `nfc`, `scan qr code`) overflow their container. Because the caption labels scale with Dynamic Type, `maxButtonWidth` grows large enough that the total width of the four buttons exceeds the screen width. 

The `ViewThatFits` fallback does not catch this correctly: it measures the layout during a first pass when `maxButtonWidth` is still `nil`, commits to `footer_standard()`, and then overflows silently after the `GeometryPreferenceReader` updates the value. As a result, the leftmost and rightmost buttons are clipped off-screen.

## Screenshots

| Before | After |
|--------|-------|
| Before | After - ScrollView |

## Fix

Detected the accessibility font size via `@Environment(\.dynamicTypeSize)`. When `isAccessibilitySize` is `true`, the `ViewThatFits` layout is replaced with a horizontal `ScrollView`. 

The scroll variant uses the exact same round buttons and `maxButtonWidth` equal-width logic — it is purely a layout change. For all non-accessibility font sizes, the standard behavior remains completely unchanged.

## How to Test
1. Launch the app in the iOS Simulator.
2. Go to Settings > Accessibility > Display & Text Size > Larger Text, and enable a large accessibility size.
3. Navigate to the Send screen. Verify that the footer buttons can now be scrolled horizontally without being clipped.

### DIFF
--- a/phoenix-ios/phoenix-ios/views/send/SendView.swift
+++ b/phoenix-ios/phoenix-ios/views/send/SendView.swift
@@ -77,6 +77,7 @@ struct SendView: View {
 	
 	@Environment(\.colorScheme) var colorScheme: ColorScheme
 	@Environment(\.presentationMode) var presentationMode: Binding<PresentationMode>
+	@Environment(\.dynamicTypeSize) var dynamicTypeSize: DynamicTypeSize
 	
 	@EnvironmentObject var navCoordinator: NavigationCoordinator
 	@EnvironmentObject var popoverState: PopoverState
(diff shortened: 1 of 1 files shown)


dpad85:
Thanks for this PR. What version of iOS did you use for the screenshots? I've tried reproducing the issue on `master` with a device iOS 16.7, a simulator iOS 17.5, and a simulator iOS 26.2, and did not manage to have the footer be clipped, even with extremely large fonts. See screenshot:

image

Vidarte-Alberto:
Thanks for the review! I managed to reproduce it using an iPhone 16 Pro simulator running iOS 18.3, iPhone 17 Pro running iOS 26.3, as well as on a physical device (iOS 26.3) running the App Store version of the app.
To reproduce it, you need to enable the "Larger Accessibility Sizes" toggle and move the slider to the right (as shown in the attached image).
Let me know if you can see it after trying this!

| Accessibility Settings |
|------------------------|
| reproduce-bug |

PR #821, opened 2026-07-27
Add Korean (ko) translation for Android and iOS

maxbitcoiner:
Adds a Korean localization covering both apps.

### Scope

| Target | Strings |
|---|---|
| `values-ko/important_strings.xml` | 333 |
| `values-ko/strings.xml` | 730 |
| `Localizable.xcstrings` | 1310 |
| `Currencies.xcstrings` | 156 |
| `InfoPlist.xcstrings` | 5 |
| `ko.lproj/about.html`, `ko.lproj/liquidity.html` | 2 files |

`ko` is registered in `knownRegions` and both HTML files are added to their variant groups.

### Following the guidelines

- `important_strings.xml` was done first, as requested
- Strings with `translatable="false"` (45) and the lorem ipsum placeholders are left untouched
- `%1$s` / `%@` / `%lld` keep their numbering; `%%` is preserved
- Apostrophes are escaped; `<b>`, `<i>` and `<a href>` markup is kept
- Newlines and leading/trailing whitespace match the source
- No keys added, removed or reordered

### Terminology

Technical terms stay in English where a Korean rendering would be less clear than the original, which is also what the French translation does: `swap`, `splice`, `BOLT`, `Tor`, `Electrum`, `LSP`, `LNURL`, `Taproot`, `Preimage`.

Terms that are well established in Korean are translated: 비트코인 (bitcoin), 라이트닝 (Lightning), 온체인 (on-chain), 시드 (seed), 채널 (channel), 유동성 (liquidity), 자동 채널 관리 (automated channel management), 복구 문구 (recovery phrase).

Android and iOS share wording wherever they share an English string, so the two apps read the same.

### Notes for reviewers

Translated with the UI context in hand: each key was matched against its usage in the Kotlin/Swift sources (screen, widget type, surrounding code) so that buttons read as buttons and error messages read as full sentences. Around 87% of strings could be traced to a call site; the rest were resolved from the key name and the developer comment.

I maintain the Korean translations for several Bitcoin projects on Transifex, so the terminology here is consistent with what Korean users already see elsewhere in the ecosystem:

- [Blink](https://app.transifex.com/blinkbtc)
- [Blixt Wallet](https://app.transifex.com/blixt-wallet)
- [BlueWallet](https://app.transifex.com/bluewallet)
- [Mempool](https://app.transifex.com/mempool)
- [Zeus](https://app.transifex.com/ZeusLN)

Happy to adjust any wording you'd prefer differently.


dpad85 (commented):
Thank you for this work, it looks good to me.

The AI reports a few inconsistencies with the capitalisation of "swap" when left in English in the Korean files. I don't know if that's relevant for Korean, so you decide if it's worth a fix. I think we can merge as-is.

phoenix-android/src/main/res/values-ko/strings.xml
```diff
@@ -0,0 +1,1044 @@
...
+    <string name="bip353_empty">아직 주소 없음…</string>
+    <string name="bip353_claim_button">내 주소 받기</string>
+    <string name="bip353_claiming">주소 받는 중…</string>
+    <string name="bip353_error">주소 받기 실패</string>
+
+    <!-- PIN code dialog -->
+
+    <string name="pincode_check_screenlock_title">잠금 PIN 입력</string>
+    <string name="pincode_check_spending_payment_title">지출 PIN 입력</string>
+    <string name="pincode_check_spending_displayseed_title">시드를 보려면 지출 PIN 입력</string>]
```
dpad85:
```suggestion
    <string name="pincode_check_spending_displayseed_title">시드를 보려면 지출 PIN 입력</string>
```

dpad85 (approved):

maxbitcoiner:
Thanks for the review, and for catching the leftover `]`.

On swap capitalisation: Korean has no case, but a few short labels put the English word first, so `swap 대기 중` sat next to `Swap 준비 완료`. I aligned those labels with the Android titles (and the matching iOS ones). Mid-sentence it stays lowercase `swap`.

jiserivas777-beep:
What's all this Gemini I don't know anything about this Korean people

PR #827, opened 2026-08-28
Add Chinese translation

jbxhh:


dpad85 (commented):
Thank you for this PR.

Are you able to also translate the `important_strings.xml` file? This file actually contains the most impactful string resources in the app, including the security warning about the seed. Those strings do need to be translated before we can merge the PR.

jbxhh:
@dpad85 Translated, please approve

PR #829, opened 2026-09-16
i18n(android): fix format-argument mismatches in 7 locales

owgreen-dev:
Two `strings.xml` resources have format arguments that don't match `values/strings.xml`, so the runtime `getString(id, …)` call and the translation disagree.

**`lnurl_pay_error_invalid_malformed`** — source is `The invoice returned by <b>%1$s</b> is malformed:\n%2$s`, but `it`, `ja`, `ko`, `pl`, `uk` stop after the first sentence and drop `%2$s`, so the actual malformed-invoice reason is never shown to those users. Restored the trailing `:\n%2$s` in each; wording otherwise unchanged.

**`electrum_dialog_cert_failure`** — source is `Failed to connect` (no arguments), but `es-419` and `pt-BR` carry `:\n%1$s`. Since the string is resolved without arguments, users see a literal `%1$s`. Dropped the stray argument.

| file | change |
|---|---|
| values-it / values-ja / values-ko / values-pl / values-uk | `lnurl_pay_error_invalid_malformed`: append `:\n%2$s` |
| values-b+es+419 / values-pt-rBR | `electrum_dialog_cert_failure`: remove `:\n%1$s` |

Found with an open-source i18n checker I maintain (shipi18n) run over the `res/` tree; each string was checked by hand against the source. Per TRANSLATION.md I've kept the translations' phrasing and only touched the format arguments — happy to adjust punctuation if a locale prefers something else.


dpad85 (approved):
Good catch, thank you for this fix. It was caused by working on an lnurl fix and the IT/PL/JP/KO/UK localisations in parallel, the translation was done on older files and not reconciled later.

owgreen-dev:
Thanks for merging, and for explaining where it came from. That reconciliation gap is exactly the thing this catches: a parallel branch changes a format string, the translations were written against the older file, key parity still looks fine, and nothing fails.

If it would be useful, the same check runs in CI offline, no API key or account:

```yaml
# .github/workflows/i18n-check.yml
name: i18n check
on: [pull_request]
jobs:
  check:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - run: >-
          npx @shipi18n/cli@2 check phoenix-android/src/main/res -s en
          --severity 'missing-key=off,untranslated=off,orphan-key=off'
```

I ran exactly that against `master` after the merge: **exit 0, all 15 locales clean**, so it goes green from day one and only fails on newly introduced drift. It reads the Android layout directly, understands `%1$s` positional arguments, and the severity flags keep it to format and plural problems rather than nagging about untranslated strings.

Separately, and unrelated to CI: the iOS side has the same class of problem in `phoenix-ios/phoenix-ios/Localizable.xcstrings`, concentrated in Arabic. I would rather open that as its own issue with the detail than bury it here.

Full scan for this repo: https://shipi18n.com/oss#phoenix

